PR #21, opened 2017-05-16
Kubernetes: Allow logdna DS pods to run on master

sajal:
The DaemonSet does not launch pods on master nodes anymore.

My env:-

Kops: 1.6.0-beta.1
Kubernetes: v1.6.2 (created using kops flag `--channel alpha`)

The DS only runs on worker nodes and skips the masters, this is because the masters have the following [taint](https://kubernetes.io/docs/concepts/configuration/assign-pod-node/#taints-and-tolerations-beta-feature):-

```
  taints:
  - effect: NoSchedule
    key: node-role.kubernetes.io/master
    timeAdded: null
```
I don't know if this taint is created by kops, or is a kubernetes(1.6.x) default. This pull request instructs the DS to ignore this taint.

### DIFF
--- a/logdna-agent-ds.yaml
+++ b/logdna-agent-ds.yaml
@@ -8,6 +8,10 @@ spec:
       labels:
         app: logdna-agent
     spec:
+      tolerations:
+      - key: "node-role.kubernetes.io/master"
+        operator: "Exists"
+        effect: "NoSchedule"
       containers:
       - name: logdna-agent
         image: logdna/logdna-agent:latest
--- a/logdna-agent-ds-master.yaml
+++ b/logdna-agent-ds-master.yaml
@@ -9,9 +9,7 @@ spec:
         app: logdna-agent
     spec:
       tolerations:
-      - key: "node-role.kubernetes.io/master"
-        operator: "Exists"
-        effect: "NoSchedule"
+      - operator: Exists
       containers:
       - name: logdna-agent
         image: logdna/logdna-agent:latest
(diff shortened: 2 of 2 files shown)


leeliu:
Thanks sajal for the PR! I'll test this with older versions to ensure the yaml doesn't choke there. A bit swamped at the moment but I'll try to get to this within a couple weeks.

leeliu:
After some testing, it seems that the yaml throws validation errors on all versions < 1.6. There is an option to disable validation but not great UX. We'll look into deciding whether to have separate yaml files for deploying on masters. I'm presuming you need it to deploy to masters to get host logs?

logdna-agent-ds.yaml
```diff
@@ -8,6 +8,10 @@ spec:
       labels:
         app: logdna-agent
     spec:
+      tolerations:
```
kylegato: I would like to propose making this even simpler.

Ideally, we should probably default to sendings logs from ALL boxes in our kubernetes cluster.

To do that, this simply needs to read:

```      
tolerations:
      - operator: Exists
```

Per the k8s docs: An empty key with operator Exists matches all keys, values and effects which means this will tolerate everything.

noqcks:
Any update on this?

I modified my ds manifest according to the suggestion of @kylegato with success and deployed it. But it would be nice to get this merged into master so I don't have to keep a local copy laying around.

leeliu:
Haven't had a chance to test the tolerations part on a version older than 1.6. Though I believe both taints and tolerations were introduced in 1.6 and there were definitely validation errors with taints that prevented the yaml from being deployed as a result on older clusters.

I think the best approach would be for a different yaml for >= 1.6 or simply a different version for needing master-based node logging. We'll likely add this as an option to https://app.logdna.com/pages/add-source > Kubernetes in the coming weeks.

leeliu:
Oops, didn't mean to rename the file. Decided to just commit it directly from comments from this pull. ff875d5fd58a868e1f10ce30a65d2da3c864bee4

PR #25, opened 2017-06-27
WIP: implement proxy support

### LINKED ISSUE #23: HTTP_PROXY
Is there a way to specify the proxy? We're behind a coorporate web proxy, and the agent doesn't seem to use the HTTP_PROXY or HTTPS_PROXY environment variables.

virth:
should resolve #23

### DIFF
--- a/lib/api-client.js
+++ b/lib/api-client.js
@@ -2,6 +2,7 @@ var debug = require('debug')('logdna:lib:api-client');
 var log = require('./log');
 var got = require('got');
 var pkg = require('../package.json');
+var tunnel = require('tunnel');
 
 module.exports.getAuthToken = function(config, agentName, socket) {
     log('Authenticating API Key with ' + config.LOGDNA_APIHOST + (config.LOGDNA_APISSL ? ' (SSL)' : '') + '...');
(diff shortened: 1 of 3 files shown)


virth:
My implementation seems to actually connect to the specified Server (in local tests) but is somehow still not working on our server with our coorperate proxy-server. I just don't know if my implementation is wrong or if the tunnel-mechanism in `api-client.getAuthToken` is just something that our proxy-server is not supporting.

The error: `Auth error:  RequestError tunneling socket could not be established, cause=getaddrinfo ENOTFOUND`

This seems to be HTTP-404 but when I try call the proxy-server with a browser I actually get a response.

Could somebody have a look at my changes and take a wild guess?

lib/api-client.js
```diff
@@ -23,7 +24,39 @@ module.exports.getAuthToken = function(config, agentName, socket) {
...
     debug(url);
     debug(postdata);
-    return got.post(url, { body: postdata, json: true, headers: { 'user-agent': config.UA } })
+
+    var proxy = process.env.http_proxy;
+    var agent;
+    if (proxy) {
+        log('Using proxy server: ' + proxy);
+        if (url.startsWith('https') && proxy.startsWith('https')) {
+            agent = tunnel.httpsOverHttps({
```
BorisWilhelms: Options object can be created once and passed to the methods.

lib/api-client.js
```diff
@@ -23,7 +24,39 @@ module.exports.getAuthToken = function(config, agentName, socket) {
...
+                proxy: {
+                    host: proxy
```
BorisWilhelms: The value passed the host property must only be the host part, without protocol and port.

phagunbaya:
What is the plan for merging this PR ?

virth:
@phagunbaya I hope I can finish this in the next two weeks

PR #50, opened 2018-03-23
Extra Parameters and Better Utilization

smusali:
Added the following options:
* -w, --winevent: set Windows Event Log Options;
* -u, --unset: clear some saved configurations;
* -l, --list: show the saved configuration;

@leeliu , review and let me know whether I need to apply utility functions for others too or the issues...


smusali:
@leeliu , all is done and ready to be merged after you review; basically, I have done the following:
- added option in order to beautifully list all configurations;
- added option in order to beautifully remove some or all configurations (or even by index);
- added option in order to configure windows event logs by provider and log names;
- updated documentation.

smusali:
@leeliu , after parsing Windows Event Logs, I am trying to send meta info as a value of `f` field where `l` is the field for line or message. Please, fix it, since api won't be able to parse - after pushing commits, I remembered `f` is for file names or app names and `f` should be `winevent` but I don't know how to send that meta info containing `source`, `providername`, `logname` and other important information.

smusali:
**TBD**: 
* if `providerName` or `logName` contains multiword string (i.e. string containing whitespace(s)), we cannot take it as a whole - need to solve this;
* combining `providerName` and `logName` values with any case (i.e. case-insensitively) of `AND` or `and` is not good solution - need to provide better solution...

smusali:
@leeliu , all discussed updates are done - you can see the changes in updated [README](https://github.com/smusali/logdna-agent/blob/cli/README.md); so, it's ready to go except one more thing - `logdna-agent-ds.yaml` should be updated.

PR #27, opened 2017-09-15
Update logdna-agent-ds.yaml

caarlos0:
The default `requests.cpu` is `100`, and the logdna-agent rarely uses that amount of CPU cores. I lowered it to 20.

```console
~
❯ kx prod-sp
Switched to context "prod-sp".

❯ k top po -l app=logdna-agent
NAME                 CPU(cores)   MEMORY(bytes)
logdna-agent-j533k   0m           40Mi
logdna-agent-qw2rr   0m           43Mi
logdna-agent-s5hgf   0m           43Mi
logdna-agent-7829l   20m          54Mi
logdna-agent-xjd67   0m           37Mi
logdna-agent-d1r1j   0m           37Mi
logdna-agent-5w5tk   20m          43Mi

~ 2s
❯ kx prod
Switched to context "prod".

~
❯ k top po -l app=logdna-agent
NAME                 CPU(cores)   MEMORY(bytes)
logdna-agent-53ltl   21m          157Mi
logdna-agent-zsmj2   0m           51Mi
logdna-agent-h7874   16m          131Mi

~
❯ kx dev
Switched to context "dev".

~
❯ k top po -l app=logdna-agent
NAME                 CPU(cores)   MEMORY(bytes)
logdna-agent-52sw4   41m          368Mi
logdna-agent-k2lkf   25m          70Mi
```

PR #51, opened 2018-04-04
set requests.cpu

caarlos0:
this is the same as #27

our master died and dns-controller wouldn't start because there were no resources because logdna was taking them.


leeliu:
20m would be 1% cpu request. It's not a limit...is that what you were looking for?

caarlos0:
> 20m would be 1% cpu request. It's not a limit...is that what you were looking for?

yes

PR #55, opened 2018-05-10
Changes in WinEvent Logging

smusali:
- Queueing in case of connection errors for `win32`;
- Removed Windows Event Log Data Parsing from agent since it is now implemented in API;
- Successfully tested on `win32` system.


lib/linebuffer.js
```diff
@@ -167,7 +168,8 @@ module.exports.setSocket = function(sock) {
         exports.resetStats();
         exclude_regex = config.exclude_regex && new RegExp(config.exclude_regex);
 
-        if (config.HTTP_RETRY && os.platform() !== 'win32') {
+        if (config.HTTP_RETRY) {
+            if (!fs.existsSync(config.FILEQUEUE_PATH)) fs.mkdirSync(config.FILEQUEUE_PATH);
```
leeliu: This line cannot be here. Linebuffer loop runs for every log line and this would kill performance.

PR #62, opened 2018-10-22
Add Dockerfile for alpine-based image

hipyhop:
I've created a Dockerfile for building the agent on the latest node 8 on alpine. I chose node 8 as there appeared to be some incompatibilities with node 10.

I've also added a yarn.lock file as the package.json does not use pinned versions and I don't believe this is a good practice.

We've only just started using this in testing so currently can't make any promises on it's stability, I just wanted to put it here in case others may find it useful.

Any feedback or suggestions would be greatly appreciated.


jakedipity:
@hipyhop This is quite an old issue and I'm not sure if you're still using the agent, but if you are and are looking for something that has better resource usage you might want to consider version 2 of the agent which is built in rust https://github.com/logdna/logdna-agent-v2

Issue #38, opened 2018-01-30
Get startup script to read file for additional parameters


klatiss:
I needed to add proxy configuration for our logdna-agent install but there wasn't an obvious place for it. I have modified the startup script to look for the file /etc/sysconfig/logdna-agent and, if there, source it's contents. Please consider adding this functionality to the packaged startup script.

```
$ diff -u logdna-agent.orig logdna-agent
--- logdna-agent.orig	2018-01-30 04:04:22.843580926 +0000
+++ logdna-agent	2018-01-30 05:12:00.456739197 +0000
@@ -20,6 +20,10 @@
 stdout_log="/var/log/$name.log"
 conf_file="/etc/logdna.conf"
 
+if [ -f /etc/sysconfig/$name ]; then
+    . /etc/sysconfig/$name
+fi
+
 get_pid() {
     cat "$pid_file"
 }
@@ -41,6 +45,7 @@
         chmod 0644 /etc/logrotate.d/logdna-agent > /dev/null 2>&1
 
         echo "Starting $name..."
+        export $PROXY
         $cmd >> "$stdout_log" 2>&1 &
         echo $! > "$pid_file"
         if ! is_running; then
```
Then I have the following content in /etc/sysconfig/logdna-agent:
`PROXY="HTTPS_PROXY=http://<address>:<port>"`

leeliu:
This looks good...do you mind turning this into a pr?

mikeadamz:
@leeliu see pr #66

PR #66, opened 2018-12-05
init script: source /etc/sysconfig/logdna-agent if it exists

mikeadamz:
This provides a standard way to source environment variables for the agent (things like HTTP_PROXY or HTTPS_PROXY) see #38


attilatar:
@mikeadamz you probably need to add a reviewer

Also, I would change the name PROXY to something like ENV_VARIABLES as it could be used for more than just proxy.

smusali:
Thanks, @mikeadamz!

PR #68, opened 2018-12-17
Adding `LOGDNA_LOGENDPOINT`

smusali:
Adding new `LOGDNA_LOGENDPOINT` environment variable which has `/logs/agent` as a default value.


smusali:
so, @leeliu , basically, 2 questions:
1. is the name `LOGDNA_LOGENDPOINT` okay? I was thinking of using `LOGDNA_LOGPATH` but it’s not that much descriptive;
2. do you wanna customize API Client Authentication Endpoint too?

PR #77, opened 2019-03-08
Update logdna-agent-v2.yaml

CJP10:


respectus:
Looks good, let's get the app name changed

PR #91, opened 2019-04-15
Support for Having Space in WinEvents

smusali:
With this, now `agent` (v1) supports `-w System,Security,Setup,Domain+Controller,"Windows+DNS",` where it takes care of the case by:
- discarding final empty argument;
- replaces `+` (makes it flexible for customers) with ` ` (whitespace);
- trimming by removing quotes as well; so,
```
> sudo logdna-agent -w System,Security,Setup,Domain+Controller,"Windows+DNS",
Windows Events: System, Security, Setup, Domain Controller, and Windows DNS have been saved to config.
```

Notes:
- Added extra flag to `utils.split` in order not to mess up with glob patterns which can be used for directories and files;
- Beautified post-config-change messages.


lib/utils.js
```diff
@@ -1,6 +1,19 @@
 // Custom Splitter:
-const split = (str, delimiter) => {
-    return (str.split(delimiter || ',').map(element => element.trim())).filter(element => element);
+const split = (str, delimiter, whitespace) => {
```
dchai76: Can we name `whitespace` better? Something that makes it clear that it's a boolean. Like `with_plus_for_whitespace`.

lib/utils.js
```diff
@@ -1,6 +1,19 @@
...
+
+    if (whitespace === undefined && typeof delimiter === 'boolean') {
```
dchai76: I dislike this - parameters should represent what they actually are. If it's being called wrong, we should fix the callers, not address wrong calls in this way.

lib/utils.js
```diff
@@ -49,13 +62,22 @@ const merge = (objects, is_json) => {
 };
 
 // Custom Processing - Combining all processes:
-const processOption = (options, config) => {
-    const newValues = merge(options.map(option => exports.split(option)));
+const processOption = (options, config, whitespace) => {
```
dchai76: Same comment about param name as above.

lib/utils.js
```diff
@@ -49,13 +62,22 @@ const merge = (objects, is_json) => {
...
+    const newValues = merge(options.map(option => exports.split(option, whitespace)));
     const oldValues = (config ? (typeof config === 'string' ? exports.split(config) : config) : []);
     const diff = newValues.filter(value => oldValues.indexOf(value) < 0).filter(value => value);
+    
+    var message = `Nothing has`;
```
dchai76: Why backticks if there's no variable interpolation?

lib/utils.js
```diff
@@ -106,10 +128,19 @@ const unsetConfig = (options, config) => {
...
+    
+    var message = 'been deleted!';
+    if (diff.length === 0) { message = `Nothing has ${message}`; }
+    else if (diff.length === 1) { message = `${diff[0]} has ${message}`; }
+    else if (diff.length === 2) { message = `${diff[0]} and ${diff[1]} have ${message}`; }
+    else {
+        const last = diff.pop();
+        message = `${diff.join(', ')}, and ${last} have ${message}`;
+    }
+
```
dchai76: This is the (virtually) the exact same logic as lines 71-77. As a hater of code duplication I'd encourage you to pull this out into a common function and (in this line) just append  `' been deleted'` at the end.

lib/utils.js
smusali: replaced it with `haswhitespace`

lib/utils.js
smusali: Updated!

lib/utils.js
smusali: Updated!

lib/utils.js
smusali: Updated! Created new function for this!

lib/utils.js
smusali: Done!

lib/utils.js
```diff
@@ -1,6 +1,22 @@
 // Custom Splitter:
-const split = (str, delimiter) => {
-    return (str.split(delimiter || ',').map(element => element.trim())).filter(element => element);
+const split = (str, delimiter, haswhitespace) => {
```
dchai76: If you don't mind, can you make it `hasWhitespace`?

lib/utils.js
```diff
@@ -49,13 +56,14 @@ const merge = (objects, is_json) => {
 };
 
 // Custom Processing - Combining all processes:
-const processOption = (options, config) => {
-    const newValues = merge(options.map(option => exports.split(option)));
-    const oldValues = (config ? (typeof config === 'string' ? exports.split(config) : config) : []);
+const processOption = (options, config, haswhitespace) => {
```
dchai76: `hasWhitespace` please

lib/utils.js
```diff
@@ -1,6 +1,22 @@
...
+            element = element.replace(/^"+/, '').replace(/"+$/, '');
+            element = element.replace(/^'+/, '').replace(/'+$/, '');
+            element = element.replace(/\+/, ' ');
+        }
+        return element.trim();
+    }).filter(element => element);
+};
+
+// Prepare Message Reporting the Options Affected:
+const preparePostMessage = (diff, length) => {
```
dchai76: I'd just pass in `diff` as the sole param and use `diff.length` where you need. You already do that in line 17 and it's a bit odd to mix that with also using a `length` param.

lib/utils.js
```diff
@@ -22,24 +38,15 @@ const stringify = (obj) => {
...
 // Custom Scaled Merger:
-const merge = (objects, is_json) => {
-    is_json = is_json || false;
-    if (is_json) {
-        return objects.reduce((merged, obj) => {
-            merged.push(Object.assign(merged.pop(), obj));
-            return merged;
-        }, [{}])[0];
-    }
+const merge = (objects) => {
```
dchai76: I'm assuming you removed these lines because `is_json` is never set to true anywhere?

smusali: yeap

lib/utils.js
smusali: Oh, I forgot to remove - gonna remove the `diff.length` from line 17 - I wanna pass `length` since I use it in multiple places - don't wanna define new variable inside function - just calculating once and passing into the function

lib/utils.js
smusali: Sure thing

smusali: Sure thing

lib/utils.js
smusali: Done!

lib/utils.js
dchai76: I'd actually much prefer the opposite - only pass in `diff` and use `diff.length` everywhere. If you *really* want (although I think it's overkill), you could do `const length = diff.length` at the top of this fn. But it's error prone to depend on callers to pass things with the right state (here that `length` reflects `diff` and here it's unnecessary.

smusali: I see - you are right - just updated!

PR #92, opened 2019-04-15
Adding CircleCI Configuration + Docker Image vulnerabilities fixes

RYadav95:
New config.yml that creates a docker image and then push it to docker hub and IBM registry.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -22,7 +22,7 @@
               root: .
               paths:
                 - ./image.tar
-      publish-tag:
+      docker-hub:
         executor: docker-publisher
         steps:
           - attach_workspace:
@@ -32,7 +32,7 @@
               name: Load archived Docker image
               command: docker load -i /tmp/workspace/image.tar
           - run:
-              name: Publish Docker Image to Docker Hub and IBM registry
+              name: Publish Docker Image to Docker Hub
               command: |
                 echo "$DOCKERHUB_PASS" | docker login -u "$DOCKERHUB_USERNAME" --password-stdin
                 IMAGE_TAG=${CIRCLE_TAG/v/''}
@@ -40,10 +40,22 @@
                 docker tag $IMAGE_NAME:latest icr.io/ext/logdna-agent:$IMAGE_TAG
                 docker push $IMAGE_NAME:latest
                 docker push $IMAGE_NAME:$IMAGE_TAG
+      ibm-registry:
+        executor: docker-publisher
+        steps:
+          - attach_workspace:
+              at: /tmp/workspace
+          - setup_remote_docker
+          - run:
+              name: Load archived Docker image
+              command: docker load -i /tmp/workspace/image.tar
+          - run:
+              name: Publish Docker Image to IBM registry
+              command: |
+                IMAGE_TAG=${CIRCLE_TAG/v/''}
                 echo "$IBM_API_KEY" | docker login -u iamapikey --password-stdin icr.io
                 docker push icr.io/ext/logdna-agent:$IMAGE_TAG
-                docker push icr.io/ext/logdna-agent:latest
-
+                docker push icr.io/ext/logdna-agent:latest                
     workflows:
       version: 2
       build-tags:
@@ -54,7 +66,15 @@
                   only: /.*/
                 branches:
                   ignore: /.*/
-          - publish-tag:
+          - docker-hub:
+              requires:
+                - build
+              filters:
+                tags:
+                  only: /.*/
+                branches:
+                  ignore: /.*/
+          - ibm-registry:
               requires:
                 - build
               filters:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -10,23 +10,26 @@
         executor: docker-publisher
         steps:
           - checkout
+          - run: echo "export DATESTAMP=$(date +%Y%m%d)" > env-vars
           - setup_remote_docker
           - run:
               name: Build Docker image
               command: |
-                docker build -t $IMAGE_NAME:latest .
+                docker build -t ${IMAGE_NAME}:latest .
           - run:
               name: Archive Docker image
-              command: docker save -o image.tar $IMAGE_NAME
+              command: docker save -o image.tar ${IMAGE_NAME}
           - persist_to_workspace:
               root: .
               paths:
                 - ./image.tar
+                - ./env-vars
       docker-hub:
         executor: docker-publisher
         steps:
           - attach_workspace:
               at: /tmp/workspace
+          - run: cat /tmp/workspace/env-vars >> $BASH_ENV
           - setup_remote_docker
           - run:
               name: Load archived Docker image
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -0,0 +1,64 @@
...
+        executor: docker-publisher
+        steps:
+          - attach_workspace:
+              at: /tmp/workspace
+          - setup_remote_docker
+          - run:
+              name: Load archived Docker image
+              command: docker load -i /tmp/workspace/image.tar
+          - run:
+              name: Publish Docker Image to Docker Hub and IBM registry
```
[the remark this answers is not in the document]
smusali: I agree with @gregswift here - `docker build`, `docker tag`, and each `docker push` should be in separate builds.

smusali:
@RYadav95, my main and biggest concern/question: where you store the credentials and the environment variables like

.circleci/config.yml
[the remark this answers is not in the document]
RYadav95: cool. I will add sysdig secure and put docker hub and ibm registry as separate jobs.

.circleci/config.yml
```diff
@@ -0,0 +1,64 @@
...
+              command: |
+                echo "$DOCKERHUB_PASS" | docker login -u "$DOCKERHUB_USERNAME" --password-stdin
```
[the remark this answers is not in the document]
RYadav95: @smusali  we can store environment variables inside CircleCI (dashboard/UI) . 
Will also make sure to use brackets next time and I will go over CircleCI Contexts as well to improve the code.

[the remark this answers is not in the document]
smusali: k, we'll revisit for `Contexts` later

.circleci/config.yml
```diff
@@ -0,0 +1,84 @@
...
+          - run:
+              name: Load archived Docker image
+              command: docker load -i /tmp/workspace/image.tar
+          - run:
+              name: Publish Docker Image to Docker Hub
+              command: |
+                echo "$DOCKERHUB_PASS" | docker login -u "$DOCKERHUB_USERNAME" --password-stdin
+                IMAGE_TAG=${CIRCLE_TAG/v/''}
+                docker tag $IMAGE_NAME:latest $IMAGE_NAME:$IMAGE_TAG
+                docker tag $IMAGE_NAME:latest icr.io/ext/logdna-agent:$IMAGE_TAG
```
gregswift: This should get moved to line 57

Dockerfile
```diff
@@ -7,6 +7,9 @@ COPY logdna.gpg /etc/
 RUN echo "deb http://repo.logdna.com stable main" > /etc/apt/sources.list.d/logdna.list && \
     apt-key add /etc/logdna.gpg && \
     apt-get -y update && \
-    apt-get -y install logdna-agent
+    apt-get -y install logdna-agent && \
+    apt-get -y install devscripts && \
+    dget -u https://launchpadlibrarian.net/405766075/systemd_229-4ubuntu21.15.dsc && \
+    dget -u https://launchpadlibrarian.net/406779380/apt_1.2.29ubuntu0.1.dsc
```
gregswift: at the end of this you'll want to include some cleanup ```\
 && rm -rf /var/lib/apt/lists/*```

I dont know what the appropriate cleanup for the `dget`

Dockerfile
```diff
@@ -7,6 +7,9 @@ COPY logdna.gpg /etc/
 RUN echo "deb http://repo.logdna.com stable main" > /etc/apt/sources.list.d/logdna.list && \
     apt-key add /etc/logdna.gpg && \
     apt-get -y update && \
-    apt-get -y install logdna-agent
-
+    apt-get -y install logdna-agent && \
+    apt-get -y upgrade apt && \
```
gregswift: It would probably be good to replace lines 11 and 12 with `apt-get -y upgrade && \`

.circleci/config.yml
```diff
@@ -0,0 +1,93 @@
...
+          - setup_remote_docker
+          - run:
+              name: Load archived Docker image
+              command: docker load -i /tmp/workspace/image.tar
+          - run:
+              name: Publish Docker Image to Docker Hub
+              command: |
+                echo "${DOCKERHUB_PASS}" | docker login -u "${DOCKERHUB_USERNAME}" --password-stdin
+                IMAGE_TAG=${CIRCLE_TAG/v/''}
+                docker tag ${IMAGE_NAME}:latest ${IMAGE_NAME}:${IMAGE_TAG}
```
gregswift: Spit and polish suggestion: Maybe try a for loop instead to reduce duplication?

```
for tag in ${IMAGE_TAG} ${IMAGE_TAG}-${DATESTAMP}; do
  docker tag ${IMAGE_NAME}:latest ${IMAGE_NAME}:${tag}
  docker push ${IMAGE_NAME}:${tag}
done
```

I'd still like to see `docker push ${IMAGE_NAME}:latest` be a step that requires an approval since we dont have testing.  I'd like to see us validate the built image in _our_ deployments before releasing it as latest.

Dockerfile
baronomasia: That may severely bloat the image and take a very long time. But we can give it a shot, why not just the 2 packages instead?

gregswift: My thought was to see what the impact was to the image, turns out it didnt really change the size at this time.  We are adding this tooling so that we can make releases happen more often to keep the image secure. If we have to adjust the package list on the regular because of new security vulnerabilities, we are just making toil for ourselves.

gregswift: Can you remove the `apt` and see if the image size stay the same? you are still restricting it here.

it should be `apt-get -y upgrade && \`

baronomasia:
Besides the question around `apt-get upgrade` vs the more specific commands for each package, I think this is good to go (pending @gregswift's suggested for loop changes).

.circleci/config.yml
smusali: in this case, he will need to do `docker push ${IMAGE_NAME}:latest` before the `for` loop. I think instead of `for` loop, comments would be better:
```
# Push with latest tag:
docker push ${IMAGE_NAME}:latest
# Tag and Push with IMAGE_TAG:
docker tag ${IMAGE_NAME}:latest ${IMAGE_NAME}:${IMAGE_TAG}
docker push ${IMAGE_NAME}:${IMAGE_TAG}
# Tag and Push with IMAGE_TAG + DATESTAMP:
docker tag ${IMAGE_NAME}:latest ${IMAGE_NAME}:${IMAGE_TAG}-${DATESTAMP}
docker push ${IMAGE_NAME}:${IMAGE_TAG}-${DATESTAMP} 
```

Dockerfile
smusali: Upgrading `apt` and `systemd` don't seem like ideal and permanent solution - what have we agreed on regarding using `alpine`? Staying on `ubuntu:16.04`?

.circleci/config.yml
gregswift: I want to have a conversation about the latest tag. will schedule something.

Dockerfile
gregswift: ubuntu updates their image monthly.  running updates in the image is generally good practice.

that being said I'd like for us to consider using the alpine image.

RYadav95: with `apt-get -y upgrade` the image size is same as specifying `apt` and `systemd`. 
This is a temp fix, we need to move to alpine.

PR #103, opened 2019-06-14
Update README.md

dchai76:
Removing bad information about using double quotes in the conf file, clarifying that glob patterns do need double quotes only in the `index.js` cli, fixing a bad URL link, and doing a language cleanup.

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,11 +9,11 @@ LogDNA's collector agent which streams log files to your LogDNA account. LogDNA 
 
 ### From an Official Release
 
-Check out the official [LogDNA site](https://logdna.com/) on how to signup for an account and get started.
+Check out the official [LogDNA site](https://logdna.com/) to learn how to sign up for an account and get started.
 
 ### From Source
 
-Follow these quick instructions to run the LogDNA agent from source:
+Follow these instructions to run the LogDNA agent from source:
 
 ```bash
 git clone https://github.com/logdna/logdna-agent.git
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -53,40 +53,41 @@ sudo node index.js -u all                                 # unset everything exc
 sudo node index.js
 ```
 
+Note that when using glob patterns with `index.js`, you must enclose the pattern in double quotes.
+
 ### Configuration File
 
-Normally a config file is automatically generated (e.g. when you set a key using `-k`), but you can create your own config file `/etc/logdna.conf` on Linux and `C:\ProgramData\logdna\logdna.conf` on Windows:
+Normally a config file is automatically generated (e.g. when you set a key using `index.js -k`) and updated (e.g. when you add a directory using `index.js -d`) but you can create your own config file `/etc/logdna.conf` on Linux and `C:\ProgramData\logdna\logdna.conf` on Windows and save your settings there:
```
smusali: Creating config files at the specified paths is not necessary because we have `-c` option for specifying the path of the config file - by default, it's `/etc/logdna.conf` (for Unix) or `C:\ProgramData\logdna\logdna.conf` (for Windows).

dchai76: I think the intent was more to indicate that they can edit the conf file at all, and since that's where it resides by default, that's what's listed. How would you phrase that intention differently?

smusali: @dchai76 

something similar to:
"""
but you can create your own config file and use it by using `index.js -c < Path to Config File >`
"""

smusali: But, @dchai76, I think mine is kinda nitpicking; therefore, I'm considering this fine!

PR #105, opened 2019-06-17
[1.5.7]: Getting Rid of `ws`

smusali:


lib/connection-manager.js
```diff
@@ -4,6 +4,7 @@ var debug = require('debug')('logdna:lib:connection-manager');
 var log = require('./log');
 var spawn = require('child_process').spawn;
 var os = require('os');
+var async = require('async');
```
dchai76: This may sound ridiculous, but alphabetically ordering `require`s within groups is helpful for maintenance.

lib/connection-manager.js
```diff
@@ -66,12 +67,63 @@ module.exports.connectLogServer = function(config, programName) {
     var proxy = getProxyFromURI(url.parse(logurl));
     if (proxy) options.agent = new HttpsProxyAgent(url.parse(proxy));
 
-    var firstrun = true;
+    // var firstrun = true;
```
dchai76: Prefer we not commit commented out code

PR #106, opened 2019-06-18
LogDNA Agent v1.6.0: The Broken Promises and The End of WebSocket

smusali:
Basically, this `Pull Request` is intended to remove `ws` and `bluebird` from the source code of `LogDNA Agent`, since we have been using `ws` for just authentication purposes and it has been causing troubles lately.

I have added simple `POST Request` to get the information about all is good to go and then running file and `winevent` log streaming.

Right now, one of the main pain points here is it is not going to pass any of our unit tests since they use `WebSocket`s.


lib/connection-manager.js
```diff
@@ -65,21 +27,44 @@ module.exports.connectLogServer = function(config, programName) {
...
-    return new Promise(resolve => {
-        debug('connecting to websocket');
-        socket.on('open', function() {
-            log('Connected to ' + config.LOGDNA_LOGHOST + ':' + config.LOGDNA_LOGPORT + ' (' + socket._socket.remoteAddress + ')' + (config.LOGDNA_LOGSSL ? ' (SSL)' : '') + (proxy ? ' via ' + proxy : ''));
-
-            if (firstrun) {
-                firstrun = false;
-
-                // start streaming logs from logdir(s) on startup
+    const socket = new WebSocket(logurl, options);
```
mikehu: I don't see a reference to this, if we are not using this, then we should probably remove this.

lib/connection-manager.js
```diff
@@ -65,21 +27,44 @@ module.exports.connectLogServer = function(config, programName) {
...
+                    , 'ENOTFOUND'
+                ].indexOf(err.code) >= 0;
+            }
+        }, (reqCallback) => {
+            return req.post(`${logurl}${config.LOGDNA_LOGENDPOINT}`, {
+                body: JSON.stringify({
+                    e: 'ls'
+                    , ls: [{
+                        e: 'l'
+                        , l: `Connected to ${config.LOGDNA_LOGHOST}:${config.LOGDNA_LOGPORT} ${(config.LOGDNA_LOGSSL ? ' (SSL)' : '')}${(proxy ? ' via ' + proxy : '')}`
```
mikehu: seems like this statement is slightly different from the actual url being used (line:49), should these be the same to reflect what url you are actually making a post to? I mean assuming that it posted to the correct location to begin with.

lib/linebuffer.js
```diff
@@ -1,24 +1,35 @@
...
+const fs = require('fs');
+const request = require('request');
+const Agent = require('agentkeepalive');
+const Queue = require('file-queue').Queue;
+
+// Internal Modules
+const log = require('./log');
+
+// Constants
+const reqheaders = {
```
mikehu: I would have this be named in the constants fashion like `REQ_HEADERS`

lib/linebuffer.js
```diff
@@ -1,24 +1,35 @@
...
+    'Content-Type': 'application/json; charset=UTF-8'
+};
+const reqheadersgzip = {
```
mikehu: Probably good to rename to `REQ_HEADERS_GZIP`

lib/connection-manager.js
```diff
@@ -65,21 +27,44 @@ module.exports.connectLogServer = function(config, programName) {
...
+                        , t: Date.now()
+                        , f: config && config.UA && config.UA.split('/')[0] || 'logdna-agent-test'
+                    }]
+                })
+                , headers: {
```
mikehu: maybe you can set this header as a constant too?

lib/connection-manager.js
```diff
@@ -65,21 +27,44 @@ module.exports.connectLogServer = function(config, programName) {
...
+                    'Content-Type': 'application/json; charset=UTF-8'
+                }
+            }, (error, response, body) => {
+                if (error) return reqCallback(error);
+                return reqCallback(null, body);
+            });
```
mikehu: this whole block, you could just make it be
```
return reqCallback(error, body);
```

lib/connection-manager.js
```diff
@@ -65,21 +27,44 @@ module.exports.connectLogServer = function(config, programName) {
...
-        debug('connecting to websocket');
-        socket.on('open', function() {
-            log('Connected to ' + config.LOGDNA_LOGHOST + ':' + config.LOGDNA_LOGPORT + ' (' + socket._socket.remoteAddress + ')' + (config.LOGDNA_LOGSSL ? ' (SSL)' : '') + (proxy ? ' via ' + proxy : ''));
-
-            if (firstrun) {
-                firstrun = false;
-
-                // start streaming logs from logdir(s) on startup
+    const socket = new WebSocket(logurl, options);
+    const req = linebuffer.setConnection(logurl, options);
```
dchai76: I said this in prior review, but let's be consistent within the file. Since we're using `var`s let's keep using them (else convert them all). Mixing `var`s and `const`s is the worst option I think.

lib/connection-manager.js
```diff
@@ -65,21 +27,44 @@ module.exports.connectLogServer = function(config, programName) {
...
+    return new Promise((resolve) => {
+        async.retry({
+            times: 5
+            , interval: 100
+            , errorFilter: (err) => {
+                return [
+                    'ETIMEDOUT'
+                    , 'ECONNRESET'
+                    , 'EHOSTUNREACH'
+                    , 'ETIMEDOUT'
```
dchai76: You listed this twice?

lib/connection-manager.js
```diff
@@ -65,21 +27,44 @@ module.exports.connectLogServer = function(config, programName) {
...
+                    , 'ESOCKETTIMEDOUT'
+                    , 'ECONNREFUSED'
+                    , 'ENOTFOUND'
+                ].indexOf(err.code) >= 0;
```
dchai76: Isn't `].includes(err.code);` more straightforward?

lib/connection-manager.js
```diff
@@ -1,56 +1,18 @@
...
+const debug = require('debug')('logdna:lib:connection-manager');
+const os = require('os');
+const url = require('url');
+const HttpsProxyAgent = require('https-proxy-agent');
+
+// Internal Modules
+const fileUtils = require('./file-utilities');
+const winUtils = require('./windows-utilities');
+const linebuffer = require('./linebuffer');
+const WebSocket = require('./logdna-websocket');
```
dchai76: Same as Mike's comment below - I think the point is to remove websocket, get rid of the reference, get rid of this line, delete `logdna-websocket.js`.

lib/connection-manager.js
smusali: Yes, it does - removed backquotes to improve readability.

lib/connection-manager.js
smusali: In terms of understanding, might be but afaik, performance-wise, `indexOf` should be faster than `includes`

lib/connection-manager.js
smusali: No, but had to make it as a constant since called in separate places.

smusali:
@mikehu @dchai76 updated! can you do second round?

Do you have suggestion regarding rewriting testing?

index.js
```diff
@@ -19,6 +19,11 @@ var connectionManager = require('./lib/connection-manager');
 var k8s = require('./lib/k8s');
 var utils = require('./lib/utils');
 
+// Variables
+var isWinAdmin; // windows only
+var processed;
```
dchai76: Personally I'd move this down closer to the block scope where it's used but up to you.

lib/config.js
```diff
@@ -1,9 +1,11 @@
-/* globals process */
 var os = require('os');
 var path = require('path');
 
 module.exports = {
     DEFAULT_LOG_PATH: os.platform() !== 'win32' ? '/var/log' : path.join(process.env.ALLUSERSPROFILE, '/logs')
+    , DEFAULT_REQ_HEADERS: { 'Content-Type': 'application/json; charset=UTF-8' }
+    , DEFAULT_HTTP_ERRORS: ['ETIMEDOUT', 'ECONNRESET', 'EHOSTUNREACH', 'ETIMEDOUT', 'ESOCKETTIMEDOUT', 'ECONNREFUSED', 'ENOTFOUND']
```
dchai76: Perhaps my last comment was unclear: you've listed `'ETIMEDOUT'` twice - intentional?

smusali: Oh, didn't notice - unintentional, of course

smusali:
@mikehu, @dchai76, I would like to ask you to review again... There are some updates...

index.js
```diff
@@ -249,38 +254,40 @@ checkElevated()
...
-    .then(dist => {
+    .then((dist) => {
         if (dist && dist.os) {
             config.osdist = dist.os + (dist.release ? ' ' + dist.release : '');
         }
-        return new Promise(resolve => {
-            request('http://169.254.169.254/latest/dynamic/instance-identity/document/', { timeout: 1000, json: true }, function(err, res, body) {
+
+        return new Promise((resolve) => {
+            request('http://169.254.169.254/latest/dynamic/instance-identity/document/', {
```
mikehu: maybe we can put this request url into a const and have a comment about what this is -- this seems like voodoo magic otherwise -- I know, not an actual change of the code, but it would be a nice improvement

lib/connection-manager.js
```diff
@@ -63,142 +81,62 @@ module.exports.connectLogServer = function(config, programName) {
...
-
-                // start streaming logs from logdir(s) on startup
-                fileUtils.streamAllLogs(config);
-
-                if (os.platform() === 'win32' && config.winevent) {
-                    if (typeof config.winevent === 'string') {
-                        config.winevent = config.winevent.split(',');
+    var req = linebuffer.setConnection(logurl, options);
+    return new Promise((resolve) => {
+        setInterval(() => {
```
mikehu: usually you want to avoid using setInterval's, also is it possible that you have a compounding problem when that async process is constantly retrying?

ie. if the service is down and you area always retrying, it's possible that you kick off many simultaneous req.post's even though all of them are failing.

lib/connection-manager.js
```diff
@@ -63,142 +81,62 @@ module.exports.connectLogServer = function(config, programName) {
...
                 }
 
-                log('Unknown event received: ' + JSON.stringify(data));
-            } else {
-                log('Unknown event received: ' + data);
-            }
-        });
+                return;
+            });
+        }, 10 * 60 * 1000); // Every 10 Minutes
```
mikehu: you might want to put this into a constant somewhere

index.js
smusali: Yeap - done!

index.js
```diff
@@ -57,10 +62,7 @@ program
     })
     .parse(process.argv);
 
-
-// windows only
-var isWinAdmin;
-
+var isWinAdmin; // windows only
```
mikehu: I think this variable's name and use can be changed.
1) `isWinAdmin` seems to suggest a boolean, but is actually a function
2) I'm not sure at what other points this would be used, but if the only point of use is on line:78, then you could just do the dynamic import there and be done with it. Otherwise I don't think it's terrible to just directly do a require instead of trying to dynamically assign it to `require('is-administrator')` -- does that make sense?

index.js
```diff
@@ -249,38 +254,40 @@ checkElevated()
...
                     config.awsid = body.instanceId;
                     config.awsregion = body.region;
                     config.awsaz = body.availabilityZone;
                     config.awsami = body.imageId;
                     config.awstype = body.instanceType;
                 }
-                resolve(macaddress.allAsync()
-                    .catch(() => {})
-                );
+                resolve(macaddress.allAsync().catch(() => {}));
```
mikehu: do you want to `reject` when there is an error?

lib/api-client.js
```diff
@@ -21,47 +27,51 @@ module.exports.getAuthToken = function(config, agentName, socket) {
...
 
-                config.authtimeout = setTimeout(function() {
-                    resolve(exports.getAuthToken(config, agentName, socket));
-                }, config.AUTHERROR_DELAY);
-                return;
+    return new Promise((resolve) => {
+        return async.retry({
+            times: 5
+            , interval: (retryCount) => {
+                return 100 * Math.pow(2, retryCount);
```
mikehu: maybe make the `100` value a constant so you can adjust it easily?
and why 100ms? do you need it to retry that quickly?

lib/connection-manager.js
```diff
@@ -60,145 +79,68 @@ module.exports.connectLogServer = function(config, programName) {
...
-                        config.winevent = config.winevent.split(',');
-                    }
-
-                    winUtils.streamEventLog({
-                        events: config.winevent
-                    }, socket);
-
+    var req = linebuffer.setConnection(logurl, options);
+    return new Promise((resolve) => {
+        setInterval(() => {
```
mikehu: I still think you will need to not run this as a `setInterval` even though it's ok to do the retry portion

index.js
smusali: `macaddress` is optional nice-to-have here; therefore, no need to `reject` in the case of `error`.

smusali:
@dchai76 @leeliu @mikehu can you review this huge Pull Request which is intended to solve `bluebird` and `websocket` issues? Thanks in advance!

P.S. All changes are in `v1.5.7` branch of `logdna/logdna-agent` - so, you can modify yourself if needed.

lib/linebuffer.js
```diff
@@ -213,12 +237,23 @@ module.exports.setSocket = function(sock) {
 
         flushtimeout = setTimeout(exports.flush, config.FLUSH_INTERVAL);
     }
+
+    return req;
```
mikehu: I feel like maybe this part can be cleaned up a little. Because in line:168 you assign `req` from this function, but then you also return it which is assigned line:151 of `connection-manager.js`. If `req` is only ever assigned inside this file (or this function) is there a need to keep it inside a closure and declare it on line:12?

Also conceptually, here's what is happening if I'm summarizing this correctly.

There is a `setConnection` function method inside `linebuffer` that creates a request object based on input properties that is then returned to `connection-manager` to hold onto for requests that it makes?

I feel like the roles/responsibilities here are a little mixed up. I would expect `linebuffer` to be only responsible for things related to the "line buffer" and if there is a `setConnection` function, that it would take in perhaps the request object to hold onto rather than being the one to create the request object. And a `setter` function that actually returns something else seems strange to me, like perhaps a true/false return would make sense, but it's like if I'm setting something, I would not expect to get something back out of it. Perhaps if you did, then you may want to call it `createConnection`, in which case it still seems to make much more sense being a function of the "connection manager" rather than a function of the "line-buffer".

index.js
```diff
@@ -295,24 +279,17 @@ checkElevated()
...
-    .then(() => {
-        debug('got auth token:');
-        debug(config.auth_token);
         debug('connecting to log server');
-        return connectionManager.connectLogServer(config, pkg.name);
-    })
-    .then(sock => {
-        socket = sock;
+        return connectionManager.connectLogServer(config, pkg.name, cb);
+    }], (error, success) => {
```
mikehu: are we swallowing the errors here?

if `os-version` fails for example, is there just a default we fallback to? or does it not matter?

smusali: Good point - normally, `os-version` shouldn't error if `agent` is on supported versions and also, currently, [in `promisified` version](https://github.com/logdna/logdna-agent/blob/master/index.js#L252-L253), it does nothing - I can just add log message saying there is such issue - also, [`connectLogServer`](https://github.com/logdna/logdna-agent/blob/v1.5.7/lib/connection-manager.js#L152) returns `null` as in `error`.

lib/connection-manager.js
```diff
@@ -1,45 +1,53 @@
...
+
+// Internal Modules
+const fileUtils = require('./file-utilities');
+const winUtils = require('./windows-utilities');
+const linebuffer = require('./linebuffer');
+const getProxyFromURI = require('./getProxyFromURI');
+const log = require('./log');
+
+// Variables
+var logurl;
```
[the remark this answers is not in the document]
smusali: We consider `logs.logdna.com` as a `host` and `443` as a port in case of `SSL` - so, it is possible to define `config.LOGDNA_LOGURL` and then use.

lib/connection-manager.js
```diff
@@ -48,157 +56,98 @@ function restartSelf() {
...
+    return async.retry({
+        times: 3
+        , interval: (retryCount) => {
+            return 50 * Math.pow(2, retryCount);
+        }
+        , errorFilter: (err) => {
+            return config.DEFAULT_HTTP_ERRORS && config.DEFAULT_HTTP_ERRORS.includes(err.code);
+        }
+    }, (reqCallback) => {
+        return req.post(logurl + config.LOGDNA_HEALTHCHECK, {
```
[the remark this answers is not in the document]
smusali: One of the reasons is I don't wanna repeat writing/using the same set of variables instead of one variable but `config.LOGDNA_LOGURL` is gonna solve this too.

index.js
mikehu: Ya, it's cool to swallow it too, just maybe we need to make a note or understand what the implications are when an error does occur (particularly for os-version since we authored that file)

lib/connection-manager.js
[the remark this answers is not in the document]
mikehu: that's true, that works too

index.js
smusali: The final `(error, success)` won't include `os-version`'s error since it's being used inside `async.waterfall`; therefore, adding `log` statement for the case of `error` in order to be able to investigate the issues much better.

lib/linebuffer.js
smusali: @mikehu, I tried to completely move `linebuffer.setConnection` into `connection-manager.createConnection` but so many things depend on it - I mean, both `file-utilities` and `windows-utilities` call `linebuffer.addMessage` which calls `linebuffer.flush` which is using that `req` from `linebuffer` - using whole `config` (assuming `config` may contain `req` as `config.req`) as a parameter/input for `flush` and/or `addMessage` doesn't make sense from the first sight or does it?

Previously, there was `linebuffer.setSocket` which was assigning globally declared `socket` inside `linebuffer`.

smusali: @mikehu, I think my recent commit is the solution to your request - created `conn` object inside `connection-manager` containing `req`, `queue`, `exclude_regex`, and `flushtimeout` information and getting passed into `file-utilities` and `windows-utilities` then into `linebuffer`.

smusali:
@mikehu, can you do final review? Testing the recent changes...

smusali:
@mikehu, I think, my last commit should resolve your concerns - can you take a look at the changes one more time? Thanks in advance!

PR #107, opened 2019-06-20
add /u01 mount to support Oracle Cloud Infrastructure's Kubernetes (OKE)

pphu:


PeterMerkert:
It's quite some time ago since I commented, but I just created a new cluster. I wanted to confirm, that this is still a solution for OKE.

c-nixon:
Hi! thanks for the contribution, unfortunately adding this to all of our manifests could cause pods to hang while creating on other clusters if the folder doesn't exist. I have added an internal tracking ticket for the issue. We may be able to address this with our helm chart.

In the mean time I'll see about getting this added to our documentation

smusali:
@pphu @PeterMerkert, thanks for the PR and suggestions!

We are not finding adding extra mount into the already-existing yaml useful since in order to support `OKE`, creating a new yaml containing this mount would be much better but we will create a new daemonset yaml for `OKE` specifically and add tests when we have the bandwidth.

Thanks again for bringing this up!

PR #109, opened 2019-06-26
Add jest and unit tests for the unsuccessful server responses. 

vilyapilya:
Add jest for unit tests and write tests for responses that are not equal to 200.


smusali:
Since the changes have recently been merged and some tests have been removed, no need for these changes yet. Therefore, closing...

PR #110, opened 2019-07-10
CircleCI monthly update bug fix + adding sysdig secure

RYadav95:


.circleci/config.yml
```diff
@@ -57,6 +55,18 @@ jobs:
           name: Publish Docker Image to Docker Hub
           command: |
             echo "${DOCKERHUB_PASS}" | docker login -u "${DOCKERHUB_USERNAME}" --password-stdin
+            TAG=${CIRCLE_TAG/v/''}
+            D_TAG="$(wget -q https://registry.hub.docker.com/v1/repositories/logdna/logdna-agent/tags -O -  | sed -e 's/[][]//g' -e 's/"//g' -e 's/ //g' | tr '}' '\n'  | awk -F: '{print $3}' | sort -g -r | head -1)"
+            lenght="$(echo $TAG | wc -m)"
+            if [ $lenght > 3 ]; then
```
smusali: why don't you just use `$(echo $TAG | wc -m)` instead of `$lenght`?

.circleci/config.yml
```diff
@@ -57,6 +55,18 @@ jobs:
...
+              IMAGE_TAG=$D_TAG
```
smusali: I'd suggest to use brackets for variables too: `${D_TAG}` instead of `$D_TAG` - this makes thing much cleaner and safer.

baronomasia:
I'm confused as to the purpose of this PR. What are we trying to here exactly?

RYadav95:
@baronomasia Monthly automatic build for LogDNA-agent was not working. For some reason git diff is not showing the change I made to fix the bug.
Added the image scanning feature as well. Using Sysdig secure for the image scan.

baronomasia:
Ah ok, I see. Where is this automatic build being deposited after it is built?

RYadav95:
So current workflow is
For a tag commit(ex. 1.5.6) circle will build a docker image. Push it to docker hub as logdna/logdna-agent:sandbox-{$DATESTAMP}. After that sysdig secure will scan the image.
After this, CircleCI will hold the pipeline for someone's approval. Once approved, CircleCI will publish the image to docker hub and ibm registry with appropriate tags.
This pipeline will also trigger at the beginning of every month.

RYadav95:
@smusali  I made all the necessary changes. Please review and merge it.

.circleci/config.yml
```diff
@@ -57,6 +55,17 @@ jobs:
           name: Publish Docker Image to Docker Hub
           command: |
             echo "${DOCKERHUB_PASS}" | docker login -u "${DOCKERHUB_USERNAME}" --password-stdin
+            TAG=${CIRCLE_TAG}
+            D_TAG="$(wget -q https://registry.hub.docker.com/v1/repositories/logdna/logdna-agent/tags -O -  | sed -e 's/[][]//g' -e 's/"//g' -e 's/ //g' | tr '}' '\n'  | awk -F: '{print $3}' | sort -g -r | head -1)"
```
gregswift: You might want to try this method. Reduces the pipes and ignores tags that shouldnt be relevant. It also restricts this to just the v1 in this repository with the MAJOR_VERSION grep.  You'll want to add this up at the top. I'll commment there too.

```D_TAG="$(wget -q https://registry.hub.docker.com/v1/repositories/logdna/logdna-agent/tags -O - | jq -r '.[] | .name' | awk '!/-/ && /\./ { print $1}' | grep ${MAJOR_VERSION} | tail -1)"```

Comment (gregswift) [reply .circleci/config.yml]: On line 13 you would want to add the `MAJOR_VERSION=1`

I also think replacing latest everywhere with a `BASE_TAG=sandbox` would probably be a good idea for simplicity and chnges in the future.

File: .circleci/config.yml [hunk 4]
```diff
@@ -57,6 +56,18 @@ jobs:
           name: Publish Docker Image to Docker Hub
           command: |
             echo "${DOCKERHUB_PASS}" | docker login -u "${DOCKERHUB_USERNAME}" --password-stdin
+            TAG=${CIRCLE_TAG}
+            D_TAG="$(wget -q https://registry.hub.docker.com/v1/repositories/logdna/logdna-agent/tags -O - | jq -r '.[] | .name' | awk '!/-/ && /\./ { print $1}' | grep ${MAJOR_VERSION} | tail -1)"d
```
gregswift: There is a `d` at the end of this line. Copy pasta or go issue I think

baronomasia: maybe it's intentional? the var name is D_TAG and this certainly would put a d character in the tag string...

PR #116, opened 2019-07-17
Allow the execution if the config file does not exist yet. 

vilyapilya:
If agent is running for the first time and the key has not been passed, it will prompt the user to pass the key. 
When the key is passed it, will be saved in the file and the execution will continue

### DIFF
--- a/index.js
+++ b/index.js
@@ -91,6 +91,11 @@ if ((os.platform() === 'win32' && require('is-administrator')()) || process.getu
             parsedConfig.key = process.env.LOGDNA_AGENT_KEY;
         }
 
+        if (!program.key && !parsedConfig.key) {
+            console.error('LogDNA Ingestion Key not set! Use -k to set or use environment variable LOGDNA_AGENT_KEY.');
+            process.exit();
+        }
+
         // allow exclude to be passed via env
         if (process.env.LOGDNA_EXCLUDE || process.env.EXCLUDE) {
             parsedConfig.exclude = process.env.LOGDNA_EXCLUDE;
@@ -117,11 +122,6 @@ if ((os.platform() === 'win32' && require('is-administrator')()) || process.getu
         }
 
         var saveMessages = [];
-
-        if (!program.key && !parsedConfig.key) {
-            console.error('LogDNA Ingestion Key not set! Use -k to set or use environment variable LOGDNA_AGENT_KEY.');
-            process.exit();
-        }
 
         if (program.key) {
             parsedConfig.key = program.key;
--- a/index.js
+++ b/index.js
@@ -262,9 +262,9 @@ if ((os.platform() === 'win32' && require('is-administrator')()) || process.getu
             return macaddress.all(cb);
         });
     }], (error, all) => {
-        if(error) {
-          console.error(error);
-          process.exit();
+        if (error) {
+            console.error(error);
+            process.exit();
         }
         if (all) {
             var ifaces = Object.keys(all);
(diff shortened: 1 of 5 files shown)


index.js
```diff
@@ -76,9 +76,13 @@ program
 
 if ((os.platform() === 'win32' && require('is-administrator')()) || process.getuid() <= 0) {
     async.waterfall([(cb) => {
-        return properties.parse(config.CONF_FILE, {
-            path: true
-        }, cb);
+        if (fs.existsSync(config.CONF_FILE)) {
```
smusali: I'd suggest to do it in the following way:
```
        return properties.parse(config.CONF_FILE, {
            path: true
        }, (error, parsedConfig) => {
            return cb(null, parsedConfig);
        })
```

which is `callback` version of [this](https://github.com/logdna/logdna-agent/blob/baf34affe3fc74bdcd16fc2e1bf0ee68daf55abe/index.js#L98-L99).

index.js
```diff
@@ -258,6 +262,10 @@ if ((os.platform() === 'win32' && require('is-administrator')()) || process.getu
             return macaddress.all(cb);
         });
     }], (error, all) => {
+        if(error) {
```
smusali: no need for error checking - because there might be errors in `distro` and `macaddress` but they shouldn't stop the execution.

index.js
mikehu: since you are already doing some other async process, it's probably better to just use the async version of `fs.exists` rather than the sync version

index.js
mikehu: spacing here, do we have lint rules turned on for this repo?
(space after `if`)

index.js
mikehu: actually ya, since you are using an `async.waterfall` here, you could just break this into a `fs.exists` step, then proceeding to `properties.parse` step.

In fact why not just convert this to a `async.seq` I think your code will read much better this way.

index.js
vilyapilya: nope, we don't

smusali: @mikehu, we don't have it in testing, it seems, but we have `.eslintrc` inside the repo - @vilyapilya, can you do `eslint --fix -c .eslintrc index.js` and push the changes? Thanks in advance!

index.js
vilyapilya: waterfall has one of the difference which is well used here is that it does  not execute the rest of the function if something fails but returns the error instead. I think this functionality is useful here. The other difference is that the async.seq takes the initial parameter as the main callback argument. This  I think will require to change  how the script is written here. But to achieve better readability, I think I should change give name to all those anonymous methods in the array and pass the names only.

smusali: @vilyapilya, I agree with @mikehu on using `fs.exists` - but [`fs.exists`](https://nodejs.org/api/fs.html#fs_fs_exists_path_callback) has already been deprecated - instead, you can use `fs.access`

vilyapilya: And fs.exists has a different callback signature.

smusali: I mean, you can use `fs.access` in the following way, if you want:
```
        return fs.access(config.CONF_FILE, (error) => {
            if (error) {
                return cb(null, null);
            }
            return properties.parse(config.CONF_FILE, {
                path: true
            }, cb);
        });
```
instead of
```
        if (fs.existsSync(config.CONF_FILE)) {
            properties.parse(config.CONF_FILE, {
                path: true
            }, cb);
        } else {
            cb(null, null);
        }
```

But it will be needed to be tested, if you add this change.

vilyapilya: replaced

smusali:
@vilyapilya, make sure you run `eslint --fix -c .eslintrc <files>` before pushing any commit

PR #117, opened 2019-07-19
v1.6.1 Changes

smusali:


index.js
```diff
@@ -76,22 +76,27 @@ program
...
+        });
     });
 };
 
 const mainExecution = (parsedConfig, cb) => {
     parsedConfig = parsedConfig || {};
 
     // allow key to be passed via env
-    if (process.env.LOGDNA_AGENT_KEY || process.env.AGENT_KEY || process.env.INGESTION_KEY) {
+    if (process.env.LOGDNA_AGENT_KEY) {
```
mikehu: if we removed this will this not be backwards compatible?

index.js
```diff
@@ -76,22 +76,27 @@ program
 
 
 const checkFileExistence = (cb) => {
-    fs.access(config.CONF_FILE, (error) => {
+    return fs.access(config.CONF_FILE, (error) => {
```
mikehu: I don't think you need `return` here, it's a bit confusing

index.js
```diff
@@ -265,46 +281,50 @@ const awsProcessing = (dist, cb) => {
...
-                process.exit();
-            }
-            if (all) {
-                var ifaces = Object.keys(all);
-                for (var i = 0; i < ifaces.length; i++) {
-                    if (
-                        all[ifaces[i]].ipv4 && (
-                            all[ifaces[i]].ipv4.indexOf('10.') === 0 ||
+    async.waterfall([
+        checkFileExistence
```
mikehu: actually, why have this encapsulate 2 different async calls instead of having those be expanded here inside the waterfall?

index.js
smusali: It's just returning function - to make sure it returns - can be used without `return` but in the future, if we add more things, we need to make sure that this is the final call to return.

index.js
smusali: Previously, we didn't have that - added [here](https://github.com/logdna/logdna-agent/commit/4b3bfd4637d4d543b824bdf6e105533b6ffc5037) then now realized it might cause some troubles - and people are used to using `LOGDNA_AGENT_KEY` (even in `k8s`) instead of others.

index.js

index.js

PR #119, opened 2019-07-29
Support for Remote Restart and Redirecting Log Ingestion to New Endpoint

smusali:


lib/connection-manager.js
```diff
@@ -54,10 +54,12 @@ const checkHealth = (config) => {
                         log('LogDNA Agent Update Request Received - Auto Updating...');
                         return utils.updateSelf();
```
dchai76: This turns out to be academic because I learned today we're not supporting auto-update in the future, but do you need to restart on an update?

smusali: Interesting question - I need (!) so, I need to change the logic little bit more...

PR #126, opened 2019-09-19
v1.6.2 Changes

smusali:
Changes include:
* `glob` fix
* using `user-agent` in `req.headers`
* Using `node v10` for `nexe` building

### DIFF
--- a/index.js
+++ b/index.js
@@ -9,10 +9,11 @@ const macaddress = require('macaddress');
 const request = require('request');
 
 // Internal Modules
-const log = require('./lib/log');
-const distro = require('./lib/os-version');
 const connectionManager = require('./lib/connection-manager');
+const distro = require('./lib/os-version');
 const k8s = require('./lib/k8s');
+const log = require('./lib/log');
+const pkg = require('./package.json');
 const utils = require('./lib/utils');
 
 // Constants
@@ -21,19 +22,8 @@ const HOSTNAME_PATH = '/etc/logdna-hostname';
 
 // Variables
 var config = require('./lib/config');
-var pkg = require('./package.json');
 var processed;
 
-// Initializations
-if (os.platform() === 'linux') {
-    pkg.name += '-linux';
-} else if (os.platform() === 'win32') {
-    pkg.name += '-windows';
-} else if (os.platform() === 'darwin') {
-    pkg.name += '-mac';
-}
-
-config.UA = pkg.name + '/' + pkg.version;
 config.CONF_FILE = program.config || config.DEFAULT_CONF_FILE;
 
 process.title = 'logdna-agent';
--- a/index.js
+++ b/index.js
@@ -247,7 +247,7 @@ if ((os.platform() === 'win32' && require('is-administrator')()) || process.getu
             }
 
             // merge into single var after all potential saveConfigs finished
-            config = Object.assign(config, parsedConfig);
+            config = Object.assign({}, config, parsedConfig);
 
             config.tags = process.env.LOGDNA_TAGS || config.tags;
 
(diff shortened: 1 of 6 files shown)


lib/connection-manager.js
```diff
@@ -36,12 +36,14 @@ const checkHealth = (config) => {
...
-                , headers: config.DEFAULT_REQ_HEADERS
+                , headers: Object.assign({}, config.DEFAULT_REQ_HEADERS, {
+                    'user-agent': config.UA
+                })
                 , qs: {
                     timestamp: Date.now()
                 }
             }, (err, res, body) => {
-                return reqCallback(err, Object.assign(JSON.parse(body || '{}'), err ? {} : {
+                return reqCallback(err, Object.assign({}, JSON.parse(body || '{}'), err ? {} : {
```
dchai76: The `{}` seems unnecessary in this case, but nbd.

lib/connection-manager.js
```diff
@@ -11,7 +11,7 @@ const fileUtils = require('./file-utilities');
 const getProxyFromURI = require('./getProxyFromURI');
 const linebuffer = require('./linebuffer');
 const log = require('./log');
-const utils = require('./log');
+const utils = require('./utils');
```
dchai76: 🤦‍♂

lib/linebuffer.js
```diff
@@ -133,7 +133,9 @@ const flush = () => {
                         url: config.LOGDNA_LOGURL + config.LOGDNA_LOGENDPOINT
                         , method: 'POST'
                         , body: out
-                        , headers: config.DEFAULT_REQ_HEADERS_GZIP
+                        , headers: Object.assign({}, config.DEFAULT_REQ_HEADERS_GZIP, {
+                            'user-agent': config.UA
```
dchai76: Am I missing something? I don't see `config.UA` defined anywhere. Let me know if I'm wrong.

smusali: @dchai76 https://github.com/logdna/logdna-agent/blob/v1.6.2/index.js#L36

dchai76: OK wow, that's confusing. I strongly suggest we move the `package` include into the `config`, define `UA` as a field there (and let's spell it out - `USER_AGENT`), and include that field specifically in `DEFAULT_REQ_HEADERS_GZIP` and `DEFAULT_REQ_HEADERS` in the config instead of augmenting here. The point of a centralized config file is to centralize config, not have it distributed in different places, so let's do that.

smusali: That's good idea! I'll keep it as `UserAgent` - consistent across all other integrations

smusali: Updated! Done!

index.js
```diff
@@ -9,10 +9,11 @@ const macaddress = require('macaddress');
 const request = require('request');
 
 // Internal Modules
-const log = require('./lib/log');
-const distro = require('./lib/os-version');
 const connectionManager = require('./lib/connection-manager');
+const distro = require('./lib/os-version');
 const k8s = require('./lib/k8s');
+const log = require('./lib/log');
+const pkg = require('./package.json');
```
dchai76: Thank you... I'm so code anal that stuff like this makes me happy

lib/config.js
```diff
@@ -1,5 +1,13 @@
...
-var path = require('path');
+// External Modules
+const os = require('os');
+const path = require('path');
+
+// Internal Modules
+const pkg = require('../package.json');
+
+// Constants
+const PKG_SYSTEMS = { linux: '-linux' , win32: '-windows' , darwin: '-mac' };
```
dchai76: I personally think this makes the next line look messy; I think it's fine to just leave the logic like it was in index.js:28-36 before.

smusali: I honestly loved this one since `LDUSERAGENT` `Environment Variable` can be used as well as a part of configuration.

PR #130, opened 2019-10-07
Fixing Configuration-related Bugs

smusali:
this covers the followings:
- being able to override `LOGDNA_LOGHOST` and/or `LOGDNA_LOGPORT` thru config file
- being able to use different config file
- ignoring to do file reading operation if it is on `Windows`
- simplifying getting `OS Distro` information
- rearranging utility functions
- moving custom logger into `./lib/utils.js`
- more informative `debug` messages


smusali:
@mikehu @dchai76 can you review this `Pull Request` if possible? Thanks in advance!

P.S. `Travis` fails due to cached build.

index.js
```diff
@@ -143,15 +142,15 @@ if ((os.platform() === 'win32' && require('is-administrator')()) || process.getu
...
 
             if (program.set && program.set.length > 0) {
-                for (var i = 0; i < program.set.length; i++) {
-                    var kvPair = utils.split(program.set[i], '=', false);
+                program.set.forEach((setOption) => {
+                    const kvPair = setOption.split('=');
                     if (kvPair.length === 2) {
                         parsedConfig[kvPair[0]] = kvPair[1];
-                        saveMessages.push('Config variable: ' + kvPair[0] + ' = ' + kvPair[1] + ' been saved to config.');
+                        saveMessages.push(`Config variable: ${kvPair[0]} = ${kvPair[1]} been saved to config.`);
```
dchai76: While we're here, format should probably be "Config variable x = y has been saved to config." (add "has").

index.js
```diff
@@ -298,7 +295,7 @@ if ((os.platform() === 'win32' && require('is-administrator')()) || process.getu
             }
         }
 
-        log(program._name + ' ' + pkg.version + ' started on ' + config.hostname + ' (' + config.ip + ')');
+        utils.log(program._name + ' ' + pkg.version + ' started on ' + config.hostname + ' (' + config.ip + ')');
```
dchai76: Use string interpolation?

lib/connection-manager.js
```diff
@@ -88,7 +87,8 @@ const checkHealth = (config) => {
 // Initiate Connection
 const connectLogServer = (config) => {
     firstrun = true;
-    debug(config.LOGDNA_LOGURL);
+    config.LOGDNA_LOGURL = `${config.LOGDNA_LOGSSL ? 'https' : 'http'}://${config.LOGDNA_LOGHOST}:${config.LOGDNA_LOGPORT}`;
```
dchai76: Same comment as above. You don't need (and shouldn't) modify the `config` object. Save `logdnaURL` as a local `const`.

index.js
```diff
@@ -65,6 +62,8 @@ program
     .parse(process.argv);
 
 if ((os.platform() === 'win32' && require('is-administrator')()) || process.getuid() <= 0) {
+    config.CONF_FILE = program.config || config.DEFAULT_CONF_FILE;
```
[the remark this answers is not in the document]
smusali: Again, this is used in `./lib/utils` at least

index.js

index.js

test/lib/getDistro.js
```diff
@@ -2,17 +2,17 @@
```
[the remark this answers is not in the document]
smusali: Done!

Gruntfile.js
```diff
@@ -18,21 +18,18 @@ module.exports = function(grunt) {
                     exclusionPattern: /browser/
```
dchai76: It's generally a bad practice to make whitespace-only changes to a file in a PR.

index.js
```diff
@@ -10,9 +10,8 @@ const request = require('request');
 
 // Internal Modules
 const connectionManager = require('./lib/connection-manager');
-const distro = require('./lib/os-version');
+const getDistro = require('./lib/getDistro');
```
dchai76: So you probably want to re-test - this doesn't work. You renamed `getDistro.js`.

smusali: Done, @dchai76!

smusali:
@mikehu can you give your final feedback after reviewing if possible? Thanks in advance!

smusali:
@dchai76 @mikehu I'd say I could let them be separate local variable without getting added to `config`

index.js
```diff
@@ -250,9 +249,7 @@ if ((os.platform() === 'win32' && require('is-administrator')()) || process.getu
                 }
             }
 
-            distro((error, dist) => {
-                return cb(null, error ? {} : dist);
-            });
+            return getDistro(cb);
```
mikehu: according to the function declaration, this expects 2 input params, `path` and `callback`, but only one is provided
upon inspection, it looks like there is some inherent contextual check around `os.platform()` so technically this will work, but this is very error-prone -- in fact based on what I can see right now, the contextual conditional check and where this code is being run seems to be in exact contradiction, which likely will just break your code

I will leave my comments around the specifics in the function declaration

lib/get-distro.js
```diff
@@ -0,0 +1,65 @@
...
+
+        return callback(null, dist);
+    });
+};
+
+module.exports = (path, callback) => {
+    if (typeof path === 'function' && os.platform() !== 'win32') {
+        callback = path;
+        path = DEFAULT_OS_PATH;
+    }
```
mikehu: so here you are expecting 2 params, but you are trying to make `path` optional, however this is only in effect when `os.platform()` value is `win32` -- even conceptually this is strange

for node, because there is technically no function overloading, you are simply sniffing input params to support multiple function signatures and so you want to be very consistent in how you shift these arguments so you don't cause code breakage

if you break this down, you are actually changing how a function signature is interpreted based on the running context of the app -- which is really bad here

(I didn't expand the entirety of index.js so I could be mistaken but...) if we reference the caller to this from line:252 in index.js, this call is nested inside an if check for `os.platform() === 'win32'` yet here, we are only shifting the path when it isn't `win32`? Which means `path` at this point would hold the callback, and so it bypasses everything, tries to call `getDist()` with undefined, and will now just throw new Error -- but since this was requested by a (likely) async function, you are now hung because the callback function (which is actually in `path`) is never called

lib/get-distro.js
```diff
@@ -0,0 +1,65 @@
...
+const log = require('./utils').log;
+
+// Constants
+const DEFAULT_OS_PATH = '/etc/os-release';
+
+// Helper in Getting Dist from GetOS
+const getDist = (callback) => {
+    if (typeof callback !== 'function') {
+        throw new Error('Must provide a function callback');
+    }
```
mikehu: you don't want to throw Errors when trying to execute an async function

in this case, you are checking if callback is actually a function, if it isn't, you should just do a no-op, sure the Error helps to identify the situation, but it doesn't help rescue you from now what is potentially a hung process

also, since this is a local function, you should be able to ensure that you are passing in the correct parameters without these safety checks -- and if you do need to ensure the params are sanitized, they should be done in the publicly exported function, not the internal private function

lib/linebuffer.js
```diff
@@ -130,7 +130,7 @@ const flush = () => {
                     level: config.GZIP_COMPRESS_LEVEL
                 }, (err, out) => {
                     config.req({
-                        url: config.LOGDNA_LOGURL + config.LOGDNA_LOGENDPOINT
+                        url: utils.buildUrl(config.LOGDNA_LOGHOST, config.LOGDNA_LOGPORT, config.LOGDNA_LOGSSL, config.LOGDNA_LOGENDPOINT)
```
mikehu: I see a lot of these utils.buildUrl(), using the same parameters over and over again, could we not add `url` to request.defaults? or does request not allow you to have a default `url`?

lib/get-distro.js
smusali: I would remove `path` forever but this function is also used in testing but I did workaround - and thanks for showing the dumb/easy failure case! Loved it!

lib/get-distro.js
smusali: Gone!

lib/linebuffer.js
smusali: I have done it - even moved `method` and `headers` too - and didn't know that you can override already-defaulted `request` fields thru [`request.defaults({...}).defaults({...})`](https://github.com/request/request#requestdefaultsoptions).

smusali:
@mikehu, changes are done! Can you take a final look over the changes? Thanks in advance! :)

smusali:
@mikehu, can you do final round of review? Thanks in advance!

PR #135, opened 2019-10-31
retry on receiving 5xx

smusali:
Retry should happen not only when it gets `http` and/or network-related exceptions but also when it receives `5xx` status code from the server which is meaning `Server Error`


weizou19:
LGTM except the test failing

lib/linebuffer.js
```diff
@@ -24,6 +24,31 @@ var dccount = 0;
 var httperr;
 var queue;
 
+// check response to see need for buffering into queue
+const checkResponse = (error, response, body) => {
```
dchai76: I would name this something like `needsRetry` - that it's returning a bool, and what that value is meant to indicate is not clear from the name.

lib/linebuffer.js
```diff
@@ -24,6 +24,31 @@ var dccount = 0;
...
+    if (error) {
+        debug('req error ' + error.code + ': ' + error);
+        if (!httperr) { httperr = Date.now(); }
```
dchai76: This precedes you so you don't have to change it but as a reader, this is pretty mystifying on a couple levels - what the var means and why it's being set to a date.

lib/get-distro.js
```diff
@@ -30,7 +30,7 @@ module.exports = (path, callback) => {
     if (os.platform() === 'win32') {
         return getDist(callback);
     }
-    
+
```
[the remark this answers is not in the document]
smusali: it happened due to `eslint --fix`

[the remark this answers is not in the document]
dchai76: Right, I recommend you only run `eslint` on modified files

lib/linebuffer.js
```diff
@@ -24,6 +24,31 @@ var dccount = 0;
 var httperr;
 var queue;
 
+// check response to see need for buffering into queue
+const checkResponse = (error, response, body) => {
+    if (error) {
+        debug('req error ' + error.code + ': ' + error);
```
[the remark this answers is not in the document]
smusali: Done! Couldn't get `error` part - can you clarify, @dchai76?

lib/linebuffer.js
smusali: Good naming - liked - done!

lib/linebuffer.js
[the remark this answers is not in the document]
smusali: Done!

lib/linebuffer.js
smusali: renaming makes it easier to understand why it's set to datetime object

smusali:
@dchai76 @weizou19 can you do second round of review?

lib/linebuffer.js
```diff
@@ -21,18 +20,43 @@ var bufferStats = {
 };
 var config = require('./config');
 var dccount = 0;
-var httperr;
+var http_error_time;
```
dchai76: `httpErrorTime`

lib/linebuffer.js
```diff
@@ -48,75 +72,56 @@ const flush = () => {
...
-            } else if (res.statusCode !== 200) {
-                debug('req failed: ' + body);
-                if (bufferStats.flushcount % 10 === 0) { utils.log('Request failed: ' + body); }
-
-            } else {
-                if (httperr) {
-                    debug('req succeeded, re-enabling');
-                    httperr = null;
-                }
+                return; // prevent fallthru
```
dchai76: `fallthrough`

smusali: done!

lib/linebuffer.js
smusali: done!

lib/linebuffer.js
```diff
@@ -21,18 +20,43 @@ var bufferStats = {
...
+            return response.statusCode >= 500;
+        } else {
+            if (httpErrorTime) {
+                debug('req succeeded, re-enabling');
+                httpErrorTime = null;
+            }
+        }
+    }
+
+    return false;
```
weizou19: I'd prefer to return as early as possible, more readable, and less error-prone:
```
if (!response) {
    return false;
}

if (response.statusCode !== 200) {
    debug(`req failed: ${body}`);
    if (bufferStats.flushcount % 10 === 0) {
        utils.log(`request failed: ${body}`);
    }
    return response.statusCode >= 500;
}

if (httpErrorTime) {
    debug('req succeeded, re-enabling');
    httpErrorTime = null;
}
return false;
```

lib/linebuffer.js
```diff
@@ -21,18 +20,43 @@ var bufferStats = {
...
-        if (Date.now() - httperr < config.HTTP_TIMEOUT / 2) {
+    if (httpErrorTime && config.TRANSPORT === 'http') {
+        if (Date.now() - httpErrorTime < config.HTTP_TIMEOUT / 2) {
             dccount = buf.length;
             return;
         } else {
-            httperr = Date.now(); // reset
+            httpErrorTime = Date.now(); // reset
             debug('req re-attempting');
         }
```
weizou19: No need for the above `else` block, as the `if` block already `return`

weizou19:
Also, do we have test case to cover the new logic?

lib/linebuffer.js
smusali: This part is done but I don't know why it doesn't show as `Outdated` - can you check this part again, @weizou19?

lib/linebuffer.js
smusali: Done!

lib/linebuffer.js
smusali: Done!

PR #137, opened 2019-11-01
update user-agent header by adding distro information

smusali:
we agreed on having `user-agent` header as in this format: `logdna-agent/1.6.2 (darwin)`, for example.

Reference:
- https://github.com/logdna/logdna-agent/blob/master/test/lib/get-distro.js#L17-L20
- https://github.com/logdna/logdna-agent/blob/master/test/assets/test-getDistro.config

### DIFF
--- a/index.js
+++ b/index.js
@@ -255,6 +255,8 @@ if ((os.platform() === 'win32' && require('is-administrator')()) || process.getu
         , (dist, cb) => {
             if (dist && dist.os) {
                 config.osdist = dist.os + (dist.release ? ' ' + dist.release : '');
+                config.DEFAULT_REQ_HEADERS['user-agent'] += ` (${dist.name && dist.name.toLowerCase() || dist.os})`;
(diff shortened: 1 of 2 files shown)


index.js
```diff
@@ -255,6 +255,8 @@ if ((os.platform() === 'win32' && require('is-administrator')()) || process.getu
         , (dist, cb) => {
             if (dist && dist.os) {
                 config.osdist = dist.os + (dist.release ? ' ' + dist.release : '');
+                config.DEFAULT_REQ_HEADERS['user-agent'] += ` (${dist.name && dist.name.toLowerCase() || dist.os})`;
+                config.DEFAULT_REQ_HEADERS_GZIP['user-agent'] += ` (${dist.name && dist.name.toLowerCase() || dist.os})`;
```
weizou19: a nit suggestion: use a temp variable:
```
const distName = ` (${dist.name && dist.name.toLowerCase() || dist.os})`;
```

smusali: Nice suggestion! Updated!

PR #138, opened 2019-11-03
autobuild for debian and redhat

smusali:
- add jobs to automate building packages for `debian` and `redhat`
- restructure `linux`-specific scripts
- update `workflow`
- update `.gitignore`

these changes are not for pushing/deploying/publishing/releasing anything - this is just for building the packages

`fpm` steps are taken from [here](https://github.com/logdna/logdna-agent/blob/master/Gruntfile.js#L24-L25)


.circleci/config.yml
```diff
@@ -4,17 +4,103 @@ executors:
...
+              --license ${LICENSE}\
+              --vendor ${VENDOR}\
+              --description ${DESCRIPTION}\
+              --url ${URL}\
+              --maintainer ${MAINTAINER}\
+              --before-remove ./.build/scripts/before-remove\
+              --after-upgrade ./.build/scripts/after-upgrade\
+              --force --deb-no-default-config-files\
+                ./.build/logdna-agent=/usr/bin/logdna-agent\
+                ./.build/scripts/init-script=/etc/init.d/logdna-agent\
```
mindjiver: We don't ship a SystemD unit file?

.circleci/config.yml
```diff
@@ -4,17 +4,103 @@ executors:
...
+              --license ${LICENSE}\
+              --vendor ${VENDOR}\
+              --description ${DESCRIPTION}\
+              --url ${URL}\
+              --maintainer ${MAINTAINER}\
+              --before-remove ./.build/scripts/before-remove\
+              --after-upgrade ./.build/scripts/after-upgrade\
+              --force\
+                ./.build/logdna-agent=/usr/bin/logdna-agent\
+                ./.build/scripts/init-script=/etc/init.d/logdna-agent\
```
mindjiver: Same here, only SysV init?

.circleci/config.yml
smusali: no - idk why but it's been like this for always

.circleci/config.yml
smusali: yea, just `init.d`

smusali:
@RYadav95 @mindjiver any extra feedback?

RYadav95:
Overall, everything looks good but I would like to know if you had any issues using two different docker images inside a single executor. Also, I think we can get rid of some code by using logdna orbs. Will submit a PR for that as soon as it's ready.

.circleci/config.yml
mindjiver: I'll dig around and see if this is something we can add, I guess Centos6 still needs a SystemV file so. Anyways. 👍

smusali:
> Overall, everything looks good but I would like to know if you had any issues using two different docker images inside a single executor. Also, I think we can get rid of some code by using logdna orbs. Will submit a PR for that as soon as it's ready.

You are right, @RYadav95! I will try to split it into 2 jobs

smusali:
@RYadav95 didn't split but separated `ruby` dependency

PR #139, opened 2019-11-07
update failure condition and debug statements

smusali:
all `2xx` response status codes indicate success and recently `207` status code was introduced regarding batches.

updated `debug` and `log` statements as well to include status code.

### DIFF
--- a/lib/linebuffer.js
+++ b/lib/linebuffer.js
@@ -34,13 +34,13 @@ const needsRetry = (error, response, body) => {
 
     if (!response) { return false; }
 
-    if (!response.statusCode.toString().startsWith('2')) {
-        debug(`[${response.statusCode}] req failed: ${body}`);
-        if (bufferStats.flushcount % 10 === 0) { utils.log(`[${response.statusCode}] request failed: ${body}`); }
+    if (!response.statusCode >= 400) {
+        debug(`[${response.statusCode}] req failed: ${JSON.stringify(body)}`);
+        if (bufferStats.flushcount % 10 === 0) { utils.log(`[${response.statusCode}] request failed: ${JSON.stringify(body)}`); }
         return response.statusCode >= 500;
     }
 
-    debug(`[${response.statusCode}] req succeeded: ${body}`);
+    debug(`[${response.statusCode}] req succeeded: ${JSON.stringify(body)}`);
     if (httpErrorTime) {
         debug('req succeeded, re-enabling');
         httpErrorTime = null;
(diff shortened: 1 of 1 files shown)


lib/linebuffer.js
```diff
@@ -34,12 +34,13 @@ const needsRetry = (error, response, body) => {
 
     if (!response) { return false; }
 
-    if (response.statusCode !== 200) {
-        debug(`req failed: ${body}`);
-        if (bufferStats.flushcount % 10 === 0) { utils.log(`request failed: ${body}`); }
+    if (!response.statusCode.toString().startsWith('2')) {
```
weizou19: This check looks uncommon. To check error response code, I usually check ```if (response.statusCode >= 400```.

lib/linebuffer.js
```diff
@@ -34,12 +34,13 @@ const needsRetry = (error, response, body) => {
...
+        debug(`[${response.statusCode}] req failed: ${body}`);
```
weizou19: Does `body` print the string we expect? I think in lots of cases, it only prints as `[object Object]`

lib/linebuffer.js
smusali: Actually, you are right! I was considering `3xx` option as well but we only reply back either `2xx` or `4xx` or `5xx` in case of `Server Error`

lib/linebuffer.js
smusali: it wasn't apparently but for making things safer, I need to stringify

lib/linebuffer.js
smusali: Updated, @weizou19!

lib/linebuffer.js
smusali: Updated, @weizou19!

weizou19: I tested on local
```
const obj = {a: 10, b: 30};
`${obj}`
```
It resolves as `[object Object]`

lib/linebuffer.js
```diff
@@ -34,12 +34,13 @@ const needsRetry = (error, response, body) => {
 
     if (!response) { return false; }
 
-    if (response.statusCode !== 200) {
-        debug(`req failed: ${body}`);
-        if (bufferStats.flushcount % 10 === 0) { utils.log(`request failed: ${body}`); }
+    if (!response.statusCode >= 400) {
+        debug(`[${response.statusCode}] req failed: ${JSON.stringify(body)}`);
+        if (bufferStats.flushcount % 10 === 0) { utils.log(`[${response.statusCode}] request failed: ${JSON.stringify(body)}`); }
```
weizou19: Should just call `stringify` once and save as local variable:
```
const bodyString = JSON.stringify(body);
```

smusali: @weizou19, nice suggestion! updated!

PR #140, opened 2019-11-10
update retry mechanism

smusali:


lib/linebuffer.js
```diff
@@ -33,15 +33,13 @@ const needsRetry = (error, response, body) => {
     }
 
     if (!response) { return false; }
-
-    const bodyString = JSON.stringify(body || {});
```
weizou19: body is likely to resolve as `[object Object]`

PR #143, opened 2019-11-16
Handle 207 better

smusali:
When `logdna-agent` sends a batch of lines containing invalid ones, `logdna-agent` can't understand `207` and in order to provide visibility over the invalid lines, this `Pull Request` is adding debug log statements to specify which log lines are invalid.

This `Pull Request` also renames some variables related to requesting and/or handling responses to make more readable.

Considering `logdna-agent` doesn't use any transport other than `http`, there is no need for `config.TRANSPORT` variable and checks.


lib/linebuffer.js
```diff
@@ -45,121 +45,123 @@ const needsRetry = (error, response, body) => {
         httpErrorTime = null;
     }
 
+    buffer = [];
```
weizou19: Why this line is needed?

smusali: to empty after each flush

dchai76: I think Wei is asking: doesn't line 70 do that already?

smusali: yup - discussed already - forgot to update: remove that part

smusali:
Updated, @dchai76 @weizou19!

dchai76:
In Phabricator land we ask for a Test Plan - how did you test this to show that the changes work. Let me ask for that here also. What did you do to test?

smusali:
> In Phabricator land we ask for a Test Plan - how did you test this to show that the changes work. Let me ask for that here also. What did you do to test?

So, @dchai76, well, this is very interesting question and there is no direct answer from me. Basically, the integration testing with test account is going to test this case.

Also, this is small addition to the code which is not affecting the current logflow and this small block was tested with the sample response body.

The other thing is `207` response code is temporary and going to be gone soon.

PR #145, opened 2019-11-26
Automate LogDNA Agent Release Pipeline

smusali:
This `Pull Request` contains new workflow called `publish.linux` and defined as following:
```
  publish-linux:
    jobs:
      - test-unit:
          filters:
            tags:
              only: /[0-9]+\.[0-9]+\.[0-9]+/
            branches:
              ignore: /.*/
      - build-linux:
          requires:
            - test-unit
          filters:
            tags:
              only: /[0-9]+\.[0-9]+\.[0-9]+/
            branches:
              ignore: /.*/
      - test-linux:
          requires:
            - build-linux
          filters:
            tags:
              only: /[0-9]+\.[0-9]+\.[0-9]+/
            branches:
              ignore: /.*/
      - deploy-linux:
          type: approval
          requires:
            - test-linux
          filters:
            tags:
              only: /[0-9]+\.[0-9]+\.[0-9]+/
            branches:
              ignore: /.*/
```


.circleci/config.yml
```diff
@@ -203,34 +148,51 @@ workflows:
...
+            - test-unit
+          filters:
+            tags:
+              only: /[0-9]+\.[0-9]+\.[0-9]+/
+            branches:
+              ignore: /.*/
+      - test-linux:
+          requires:
+            - build-linux
+          filters:
```
RYadav95: Do we need to repeat filters if we're already mentioning that `test-linux` build requires `build-linux`?

smusali: we may not but for keeping things safe and there is no harm in doing this, I'd say let's keep them

RYadav95: hrmmm I don't see any harm in removing it but yeah for now I'll approve changes.

PR #153, opened 2020-01-16
update building image pipeline

smusali:
There has been a bug in building and releasing Docker image - this `Pull Request` is intended to fix


.circleci/config.yml
```diff
@@ -7,6 +7,11 @@ executors:
       MAJOR_VERSION: 1.
     docker:
       - image: circleci/buildpack-deps:stretch
+tagged_build_filters: &tagged_build_filters
```
RYadav95: This is pretty cool. Is this `type` defined in CircleCI? Wasn't able to find any documentation.

smusali: @RYadav95, I have already tested that pattern on some others - added with @mindjiver's suggestion

RYadav95: Yeah!! I was just wondering where you guys got this idea from. If I can read a little more about this for my own knowledge

mindjiver: I'll hook you up with some docs tomorrow.

PR #157, opened 2020-03-03
[v1.6.3]: Fix Memory Leak and Improvements

smusali:
This `Pull Request` contains lots of version dependency updates mainly because of the issues and warnings their older versions had.

According to [this issue](https://github.com/npm/inflight/issues/5), [`inflight`](https://www.npmjs.com/package/inflight) had a memory leak issue and was being used by [`glob`](https://www.npmjs.com/package/glob) which is among our dependencies.

I have also converted `mocha` tests to `jest` to make it easy to extend for our developers, replaced the usage of [`macaddress`](https://www.npmjs.com/package/macaddress) with built-in `os` and also cleaned up `Gruntfile.js` too.


index.js
```diff
@@ -255,42 +254,32 @@ if ((os.platform() === 'win32' && require('is-administrator')()) || process.getu
...
 
+        const networkInterface = Object.values(os.networkInterfaces()).reduce((networkData, interfaces) => {
+            interfaces.forEach(interface => networkData.push(interface));
+            return networkData;
+        }, []).filter((interface) => {
+            return interface.family && interface.family === 'IPv4' && interface.mac && interface.address && (
+                interface.address.startsWith('10.') ||
+                interface.address.startsWith('172.') ||
+                interface.address.startsWith('192.168.'));
+        })[0];
```
matt-march: Does this mean we're changing from picking the last interface given to the first?

Also, we're not saving ipv6 addresses anymore?

smusali: if you take a look at the current implementation [here](https://github.com/logdna/logdna-agent/blob/master/index.js#L276-L291), you can see in the `if` condition we just check for the existence of `ipv4` despite of taking `all[ifaces[i]].ipv4 || all[ifaces[i]].ipv6` - we have always taken the `ipv4` if possible.

matt-march: Ah, good call about ipv6. Saw that and then promptly forgot it.

.circleci/config.yml
```diff
@@ -24,7 +24,7 @@ jobs:
     steps:
       - checkout
       - run: npm install
```
Foxyboo, line 26: [](https://app.snyk.io/invite/link/accept?invite=65497bf5-7698-4186-8111-3b3210c7b7c9&utm_source=link_invite&utm_medium=referral&utm_campaign=product-link-invite&from=link_invite)

index.js
```diff
@@ -3,7 +3,6 @@ const async = require('async');
 const debug = require('debug')('logdna:index');
 const fs = require('fs');
 const getos = require('getos');
-const macaddress = require('macaddress');
```
Foxyboo, line 6: [](https://app.snyk.io/invite/link/accept?invite=65497bf5-7698-4186-8111-3b3210c7b7c9&utm_source=link_invite&utm_medium=referral&utm_campaign=product-link-invite&from=link_invite)

lib/linebuffer.js
```diff
@@ -171,6 +171,10 @@ const flush = () => {
 };
 
 const addMessage = (message) => {
+    // ignore empty message
+    if (message.l === '') {
+        return;
```
Foxyboo, line 176:
```suggestion
-         return;_[[@]()]()_
```
]()

package.json
```diff
@@ -6,9 +6,9 @@
   "scripts": {
     "build.linux": "bash ./scripts/linux/build.sh",
     "deploy.linux": "bash ./scripts/linux/deploy.sh",
-    "lint": "./node_modules/.bin/grunt eslint",
+    "test.lint": "./node_modules/.bin/eslint -c .eslintrc *.js **/*.js",
```
Foxyboo, line 9: - [ ] #- [~~[]()~~]()

Foxyboo (commented):
#231

PR #159, opened 2020-03-06
Add slack notification agent

emily2005:
This PR is for https://app.asana.com/0/1161519493631887/1164792207792569/f.

Rohit helped with some testing but once it was added to a testing location in my branch i was able to see the slack notification. I attached a screen shot. Naturally when this runs it will be from master and from the circleci bot not my user. Please look it over and make sure it is ok! 

image

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -236,6 +236,12 @@ workflows:
       - docker-sysdig-scan:
           requires:
             - image-build
+      - slack/approval-notification:
+          webhook: ${SLACK_WEBHOOK}
+          message: "Deploy pending approval: https://circleci.com/workflow-run/$CIRCLE_WORKFLOW_ID"
+          channel: "CU50G97PW" # Optional: If set, overrides webhook's default channel setting
+          requires:
+          - docker-sysdig-scan
       - approve-docker-image:
           type: approval
           requires:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -228,6 +228,12 @@ workflows:
           filters:
             branches:
               only: master
+      - slack/approval-notification:
+          webhook: ${SLACK_WEBHOOK}
+          message: "Deploy pending approval: https://circleci.com/workflow-run/$CIRCLE_WORKFLOW_ID"
+          channel: "CU50G97PW" # Optional: If set, overrides webhook's default channel setting
+          requires:
+          - docker-sysdig-scan
     jobs:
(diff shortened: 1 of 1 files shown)


RYadav95:
You should add the same workflow to monthly build as well

.circleci/config.yml
```diff
@@ -199,6 +201,12 @@ workflows:
           requires:
             - image-build
           filters: *tagged_build_filters
+      - slack/approval-notification:
+          webhook: ${SLACK_WEBHOOK}
+          message: "Deploy pending approval: https://circleci.com/workflow-run/$CIRCLE_WORKFLOW_ID"
```
smusali: Can you change the notation from `$CIRCLE_WORKFLOW_ID` to `${CIRCLE_WORKFLOW_ID}` in order to avoid confusion?

.circleci/config.yml
```diff
@@ -199,6 +201,12 @@ workflows:
...
+          channel: "CG2K78UH2" # Optional: If set, overrides webhook's default channel setting
```
smusali: Can we hide `channel` ID? We can add it to the list of env variables as `SLACK_CHANNEL`

emily2005:
> You should add the same workflow to monthly build as well

I added it to the monthly build I think! let me know how it looks!

.circleci/config.yml
emily2005: @smusali please make sure i did this right! thank you so much

.circleci/config.yml
emily2005: @smusali do I need to have the channel id here at all? Just curious.

emily2005:
Thank you guys so much for reviewing this and making sure its right! I appreciate it.

.circleci/config.yml
smusali: @emily2005, not sure but you are using `channel: "CG2K78UH2"` here as `channel: "<channel ID>"`, right? And it's working; so, just hiding is gonna be enough.

smusali: @emily2005, I still see the channel ID explicitly here - can we use environment variable here like `SLACK_CHANNEL`? CC: @gregswift

emily2005: @smusali i added the variable in circleci and replaced it in the code. Please let me know if this is ok!

emily2005:
Samir said it is gtg and in Agent we can merge our own PR's.

PR #162, opened 2020-03-20
Add a monthly-build-manual job to enable testing

mindjiver:
Since there is no way to trigger a cron scheduled build manually lets use the instructions
from https://blog.wurbanski.me/posts/2019-10-04-circleci-manual-scheduled-workflows/ to setup a workflow that re-uses the definition from the monthly job but only triggers if a certain parameter is set in the payload sent to the CircleCI API. This way we can use curl to POST in
the neccessary parameters and that way test the cron based workflow without having to wait
for the cron expression to trigger.

How to verify that this works:
```shell
$ curl -X POST --header "Content-Type: application/json" -d '{
    "branch":"add-a-manual-job-to-trigger-monthly-builds",
    "parameters": {
      "run_monthly_build": true
    }
  }' https://circleci.com/api/v2/project/github/logdna/logdna-agent/pipeline?circle-token=${CIRCLE_API_USER_TOKEN}
```

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -17,6 +17,10 @@ executors:
       MAJOR_VERSION: 1.
     docker:
       - image: circleci/buildpack-deps:stretch
+parameters:
+  run_monthly_build:
+    type: boolean
+    default: false
 orbs:
   slack: circleci/slack@3.4.2
 jobs:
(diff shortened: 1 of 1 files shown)


emily2005:
Going forward tho this is how we should absolutely test workflows in circleci instead of creating tags. I am going to leave this PR open a bit in case anyone wants to review it in #golden-path

PR #166, opened 2020-04-15
Update logdna-agent-v2-beta.yaml

CJP10:


CJP10:
This will update the yaml to be inline with that we are using internally.

Issue #168, opened 2020-04-30
journald logs not showing systemd service name as app name


summatix:
I'm setting up LogDNA with Fedora CoreOS and am using the logdna-agent's `USEJOURNALD` feature. The logs are being forwarded correctly but aren't being correctly identified inside the LogDNA view.

Logs are showing up with the `App` "podman" instead of showing the systemd service name.

This is due to this line here: https://github.com/logdna/logdna-agent/blob/master/lib/file-utilities.js#L215

The parsing is grabbing `SYSLOG_IDENTIFIER` first instead of `_SYSTEMD_UNIT`.

smusali:
Resolved by #210 and will be released as a part of `v2.1.0`

PR #169, opened 2020-04-30
Use systemd service name as the app name

summatix:
https://github.com/logdna/logdna-agent/issues/168


summatix:
I've built this and have it deployed. Logs are showing the correct name now in my LogDNA views.

jimaek:
Are there any problems with this PR? I would love to see it merged.

smusali (approved):

jimaek:
I believe it will fix the issue below right?
docker

Currently all logs are show syslog as the reporting app which makes them very hard to read.

c-nixon:
This looks good to me and I think it's the correct behaviour, my only question is whether it could break any existing code out there that relies on the current behaviour. I think this might be a major semver change in that case?

@darinspivey @smusali Does the next release have any backcompat guarantees that would mean we need to make this opt in?

darinspivey:
> my only question is whether it could break any existing code out there that relies on the current behaviour.

Agreed.  I have a few concerns with merging this change:

1. We have to stop doing (and contributing) to these long conditionals.  They're hard to test, and impossible to change safely (as we're seeing from the difficulty presented here).

2. Although this change works for this particular use case, we don't know who out there might break from it.  There could be people who have already navigated this conditional to make values what they need.

3.  There is no way to make this "backwards compatible" code without adding an opt-in as @c-nixon suggested, and we're trying to archive this project, not add new features.

We need to loop in some more people such as product to make this decision, but my vote is to not take the risk, close it, and move forward fixing this in places where we will have more release control (such as the brand new repos/releases in the works)

darinspivey (approved):
Approving this change since the public has requested it.  It hopefully won't "break" anyone, but we can solve that problem if/when it arises.  At worst, it'll just munge the expected `app` value which isn't terrible.

smusali:
@summatix, this `Pull Request` has some merge conflicts; can you rebase this on top of `master`? Thanks!

smusali:
Been done by #210 - thanks!

PR #170, opened 2020-05-01
Added containers to build binaries

summatix:
This makes building binaries a lot more simple and consistent.

I also updated nexe to use a pre-compiled version of nodejs, so it doesn't need to be built from source. This significantly reduces the build time.

Packaging hasn't been updated because the current instructions in `CONTRIBUTING.md` are not working (I believe because `Gruntfile.js` references non-public files in the repository).

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -3,3 +3,6 @@
 .dockerignore
 .git
 Dockerfile
+logdna-agent
+node_modules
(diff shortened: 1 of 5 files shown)


smusali:
Thanks, @summatix, for the PR but we have already created plan for automating the building steps in different ways!

Issue #167, opened 2020-04-29
Windows: EBUSY: resource busy or locked


summatix:
I'm setting up logdna-agent on a Windows machine for a log file that is continuously written to. The program will often open the file and write to it. The writing causes a lock which interferes with reading the file's contents.

Unfortunately this leads to logdna-agent failing with the error `"0700 [error] tail error: [FILE]: Error: EBUSY: resource busy or locked, open '[FILE]'`.

I can replicate this in PowerShell using `Get-Content [FILE] -Wait`. This will stream contents until the program generating the log locks the file, causing the PowerShell command to fail.

I would expect logdna-agent to not throw an error when the file is locked and instead wait until the file isn't locked and the contents can be read.

summatix:
The error seems to be getting thrown here: https://github.com/logdna/logdna-agent/blob/master/lib/file-utilities.js#L134

I believe what is happening is that the program is actually erasing all previous contents of the log and then writing to the same file again.

rmcto:
I agree with summatix.

summatix:
Maybe adding something like the following to `TailReadStream._handleError` could work:

    if (error.code === "EBUSY") {
        debug(`${this._filepath}: file locked, waiting for it to be released...`);
        this._idle += this._interval;
        this._retryInInterval();
        return;
    }

summatix:
Just tested, and what I posted above does indeed work. I'll create a PR.

darinspivey:
Hey guys, sorry this is a late reply, but we're doing some re-organization of this project and addressing some bugs.  On that note, we have completely re-written the existing "tail" logic to be a lossless, fault tolerant implementation.  Stay tuned soon for the announcement of what will be called `@logdna/tail-file` (we are open sourcing this solution for the public).

For now, in an effort to clean up these issues before doing some re-organizing, I'm going to go ahead and close this, but we certainly appreciate the discussion around this.  If this is still a problem with the new implementation, the suggested fix to tolerate `EBUSY` can be easily implemented.

PR #171, opened 2020-05-01
Retry reading when file is locked

summatix:
https://github.com/logdna/logdna-agent/issues/167

This fixes a bug in the Windows version when applications have the file open and obtain a lock. The solution is simply to wait until the process has released the lock and read again.

I've tested this and the logs are showing up currently in LogDNA.


smusali (changes requested):
Thanks for the PR, @summatix! Can you please remove the semi-colons from your changes?

image

Thanks!

darinspivey:
@summatix it's also worth noting that the underlying `TailReadStream` will be replaced soon by some new code that we authored.  When its initial version is merged into the main branch, we will certainly add this support to it (and share it with you and the world, of course!)  A constructor option that takes a "rery" list of codes and/or messages would keep it easy to use and generic.  Standby!

smusali (approved):

smusali:
Done in [here](https://github.com/logdna/logdna-agent/pull/208/commits/e4a2d6caf2cf26098343d27f2017b71eb952959c)

PR #179, opened 2020-06-10
deprecation: remove WinEvent Logging Support

smusali:
After building and enabling WinEvent Logging Support, we have seen some issues in CPU Usage and compatibility. Besides that, we have also seen the lack of possibility of streaming logs from the remote locations and we have started offering NXLog as a solution; so, that means there is no more need for WinEVent Logging Support.

Semver: minor


smusali:
Postponed and can be closed

PR #180, opened 2020-06-10
lint: update eslint dependency

smusali:


smusali:
@matt-march, @jakedipity, can you take a look at this? Thanks!

jakedipity:
Is our @answerbook publicly available? If not we might need to reconsider this.

lib/welp.js
```diff
@@ -45,7 +45,7 @@ CustomWELR.prototype._cleanLogString = function(logString) {
 
 CustomWELR.prototype._checkLogString = function(logString, logName) {
     try {
-        var parsedJSON = JSON.parse(this._cleanLogString(logString));
+        let parsedJSON = JSON.parse(this._cleanLogString(logString));
```
jakedipity, line 48: try catch

lib/tailreadstream/line-splitter.js
```diff
@@ -3,9 +3,9 @@
 
 // https://nodejs.org/api/stream.html#stream_object_mode
 
-var StringDecoder = require('string_decoder').StringDecoder;
-var Transform = require('stream').Transform;
-var util = require('util');
+let StringDecoder = require('string_decoder').StringDecoder;
+let Transform = require('stream').Transform;
+let util = require('util');
```
jakedipity, line 8: Would `const` be better here?

index.js
```diff
@@ -19,8 +19,8 @@ const HOSTNAME_IP_REGEX = /[^0-9a-zA-Z\-.]/g;
 const HOSTNAME_PATH = '/etc/logdna-hostname';
 
 // Variables
-var config = require('./lib/config');
-var processed;
+let config = require('./lib/config');
```
jakedipity, line 22: Would this be better as a `const`?

index.js
```diff
@@ -166,10 +166,10 @@ if ((os.platform() === 'win32' && require('is-administrator')()) || process.getu
                 if (typeof program.list === 'boolean') {
                     program.list = ['all'];
                 }
-                var conf = properties.parse(fs.readFileSync(conf_file).toString());
+                let conf = properties.parse(fs.readFileSync(conf_file).toString());
```
jakedipity, line 169: this can be a `const`

index.js
```diff
@@ -166,10 +166,10 @@ if ((os.platform() === 'win32' && require('is-administrator')()) || process.getu
...
                 const listResult = utils.pick2list(program.list, conf);
                 if (listResult.valid) {
-                    var msg = utils.stringify(listResult.cfg);
+                    let msg = utils.stringify(listResult.cfg);
```
jakedipity, line 172: this can be a `const`

index.js
```diff
@@ -231,7 +231,7 @@ if ((os.platform() === 'win32' && require('is-administrator')()) || process.getu
                 });
             }
 
-            // merge into single var after all potential saveConfigs finished
+            // merge into single let after all potential saveConfigs finished
```
jakedipity, line 234: I think it would be more appropriate to refer to the variables as `var` instead of let in a comment :)

lib/file-utilities.js
```diff
@@ -35,17 +35,17 @@ const globalExclude = [
...
-var files = [];
-var tails = [];
+let firstrun = true;
+let files = [];
+let tails = [];
 
 const getFiles = (config, dir, callback) => {
     // glob patterns always use / (even on windows)
-    var globdir = dir.replace('\\', '/');
+    let globdir = dir.replace('\\', '/');
```
jakedipity, line 44: This can be a `const`

lib/file-utilities.js
```diff
@@ -94,7 +94,7 @@ const getFiles = (config, dir, callback) => {
 
 const streamFiles = (config, logfiles, callback) => {
     logfiles.forEach((file) => {
-        var tail, i, labels;
+        let tail, i, labels;
```
jakedipity, line 97: We don't need to forward declare `tail` and `i`. Let's just declare them when we need them.

lib/file-utilities.js
```diff
@@ -165,7 +165,7 @@ const streamAllLogs = (config, callback) => {
                 debug(logfiles);
 
                 // figure out new files that we're not already tailing
-                var diff = logfiles.filter(element => files.indexOf(element) < 0);
+                let diff = logfiles.filter(element => files.indexOf(element) < 0);
```
jakedipity, line 168: This can be a `const`

lib/file-utilities.js
```diff
@@ -194,7 +194,7 @@ const streamAllLogs = (config, callback) => {
 
     if (config.usejournald && firstrun) {
         log(`streaming from journalctl: ${config.usejournald}`);
-        var journalctl, lastchunk, i;
+        let journalctl, lastchunk, i;
```
jakedipity, line 197: Let's fix this forward declaration too

lib/getProxyFromURI.js
```diff
@@ -11,23 +11,23 @@ function formatHostname(hostname) {
...
     zone = zone.trim().toLowerCase();
 
-    var zoneParts = zone.split(':', 2);
-    var zoneHost = formatHostname(zoneParts[0]);
-    var zonePort = zoneParts[1];
-    var hasPort = zone.indexOf(':') > -1;
+    let zoneParts = zone.split(':', 2);
+    let zoneHost = formatHostname(zoneParts[0]);
+    let zonePort = zoneParts[1];
+    let hasPort = zone.indexOf(':') > -1;
```
jakedipity, line 17: These can all be `const`.

Also we don't need to a string operation to determine `hasPort`, we can just look at the length of the `zoneParts`:

```
const hasPort = zoneParts.length == 2;
```

smusali:
> Is our @answerbook publicly available? If not we might need to reconsider this.

I set the `GITHUB_PACKAGES_TOKEN` environment variable to be used by `CircleCI` - considering the environment variables are not public here, is there any other concern? + it's saved as a part of `devDependencies`; so, whenever we release a new version, we should do `npm install --production` which doesn't include `eslint`, etc.

index.js
smusali: No, @jakedipity, it gets reassigned in the bottom - I think I tried to make it `const` before but failed every time since it requires some rearrangements of the lines

lib/welp.js
smusali: It's already inside `try...catch`

lib/tailreadstream/line-splitter.js
smusali: Done!

index.js

index.js

index.js
smusali: My bad - had to exclude the comments - done!

lib/file-utilities.js
smusali: Done!

lib/file-utilities.js
smusali: Done!

lib/file-utilities.js
smusali: Done!

lib/file-utilities.js
smusali: Done!

lib/getProxyFromURI.js
smusali: Done!

smusali:
@jakedipity, can you do another round of review here, please, if possible? Thanks!

package.json
```diff
@@ -30,12 +30,22 @@
     "request": "^2.88.2"
   },
   "devDependencies": {
-    "eslint": "^6.8.0",
+    "@answerbook/eslint-config-logdna": "^4.0.0",
```
darinspivey, line 33: this is at 5.0.0. now.  Might as well make it current.

index.js
darinspivey: oof, it's *really* weird to mutate something that came from `require`.  Can you re-work it?

lib/file-utilities.js
```diff
@@ -35,17 +35,17 @@ const globalExclude = [
 ];
 
 // Variables
-var firstrun = true;
-var files = [];
-var tails = [];
+let firstrun = true;
+let files = [];
+let tails = [];
```
darinspivey, line 40: These get re-assigned?  As long as you've gone through all of these to check for `const` capability.

darinspivey:
Commit titles for dependency updates should read like this:
```
deps: @answerbook/eslint-config-logdna@5.0.0
```

matt-march:
> > Is our @answerbook publicly available? If not we might need to reconsider this.
> 
> I set the `GITHUB_PACKAGES_TOKEN` environment variable to be used by `CircleCI` - considering the environment variables are not public here, is there any other concern? + it's saved as a part of `devDependencies`; so, whenever we release a new version, we should do `npm install --production` which doesn't include `eslint`, etc.

I imagine anyone that's not part of @answerbook will get a bunch of errors if the clone the repo, wouldn't they?

darinspivey:
> > > Is our @answerbook publicly available? If not we might need to reconsider this.
> > 
> > 
> > I set the `GITHUB_PACKAGES_TOKEN` environment variable to be used by `CircleCI` - considering the environment variables are not public here, is there any other concern? + it's saved as a part of `devDependencies`; so, whenever we release a new version, we should do `npm install --production` which doesn't include `eslint`, etc.
> 
> I imagine anyone that's not part of @answerbook will get a bunch of errors if the clone the repo, wouldn't they?

Yep, true, we should not have any private packages in here.  We could probably make our linter public-facing, but I don't think we have a public registry properly set up for that yet (it's a bit out of scope here).  At this point, we could just update the instructions to tell people to `npm install --production`, which would skip that dev dependency.  They probably wouldn't be able to run tests though.  Not sure what's best.

PR #185, opened 2020-08-12
fix(networkInterface): make sure to check `networkInterface` is defined

smusali:
This is one of the potential bugs, at least, one of the customers has already reported. The test coverage is going to be added later in different Pull Request.

Semver: patch

### DIFF
--- a/index.js
+++ b/index.js
@@ -277,8 +277,10 @@ if ((os.platform() === 'win32' && require('is-administrator')()) || process.getu
                 interface.address.startsWith('192.168.'));
         })[0];
 
-        if (networkInterface.mac) { config.mac = networkInterface.mac; }
-        if (networkInterface.address) { config.ip = networkInterface.address; }
+        if (networkInterface) {
+            if (networkInterface.mac) { config.mac = networkInterface.mac; }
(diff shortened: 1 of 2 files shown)


index.js
```diff
@@ -277,8 +277,10 @@ if ((os.platform() === 'win32' && require('is-administrator')()) || process.getu
                 interface.address.startsWith('192.168.'));
         })[0];
 
-        if (networkInterface.mac) { config.mac = networkInterface.mac; }
-        if (networkInterface.address) { config.ip = networkInterface.address; }
+        if (networkInterface) {
```
darinspivey, line 280: This is fine to fix the ref error, but I think we have to look at this more closely in later revisions.  Machines can have multiple NICs, we don't do anything with ipv6 (maybe we don't need to tho), and the IP check is pretty assumptive.  I wonder if this could be a question to the installer, and defaulted to this logic.

smusali: @darinspivey, you are right! Similar to what we have found on the Node.js Code Library, we weren't capturing IPv6 but in the later revisions we can add this as well to the refactor and improvements.

PR #186, opened 2020-08-14
test: Add `tap` as the test framework along with some starter tests

darinspivey:
**test: Add `tap` as the test framework along with some starter tests**

We are moving to `tap` as a testing framework.  This commit converts
the `getFiles()` test from Jest to tap, and adds the start of more
tests.  The goal will be to eventually have some integration tests here.

Semver: patch
Ref: LOG-6026

---

**style(linter): Change linter styles**

This changes aligns with our other repos, namely a 2-space
indent.

Semver: patch
Ref: LOG-6026


darinspivey:
Closing this in favor of a new repository name and package namespace.

index.js
```diff
@@ -1,301 +1,314 @@
...
-    })
-    .parse(process.argv);
-
-if ((os.platform() === 'win32' && require('is-administrator')()) || process.getuid() <= 0) {
-    let conf_file = program.config || config.DEFAULT_CONF_FILE;
-    debug(`Path to Configuration File: ${conf_file}`);
+var config = require('./lib/config')
+var processed
+
+commander.title = 'logdna-agent'
```
jakedipity, line 27: This should remain as `process.title = 'logdna-agent'`

index.js
```diff
@@ -1,301 +1,314 @@
...
+commander._name = 'logdna-agent'
```
jakedipity, line 28: According to the documentation and code, the correct way to set a name for the commander program is to grab the `Commander` class from the module and instantiate a new object:

```
const { Command } = require('commander')
const program = new Command('logdna-agent')
```

test/unit/file-utilities.js
```diff
@@ -0,0 +1,150 @@
...
+    fileUtilities.getFiles({}, '/tmp/*.something', (err, files) => {
+      tt.type(err, Error, 'Expected error received')
+      tt.match(err, {
+        name: 'Error'
+      , message: 'mock fs.readdir error'
+      })
+      tt.end()
+    })
+  })
+})
```
jakedipity, line 150: Might be worth adding tests for symlinks and hardlinks?

index.js
jakedipity: Link to docs: https://nodejs.org/api/process.html#process_process_title

index.js
jakedipity: https://github.com/tj/commander.js/#declaring-program-variable

index.js
darinspivey: `process` is now relegated to the added function for processing command line args versus direct calling it with a test.  `commander.title` is still the `Commander` instance, so I'm not sure what you mean here?

test/unit/file-utilities.js
darinspivey: Sure, I can put a symlink in there

index.js
darinspivey: I was trying to change as little of this as possible.  Not sure what the difference between your suggestion and `const commander = require('commander')` is...

index.js
darinspivey: Fixed per the docs.  Thanks!

index.js
darinspivey: Fixed per the docs

test/unit/file-utilities.js
darinspivey: done!

package.json
```diff
@@ -4,8 +4,9 @@
   "description": "LogDNA Collector Agent",
   "main": "index.js",
   "scripts": {
-    "lint": "./node_modules/.bin/eslint -c .eslintrc *.js **/*.js",
-    "unit": "./node_modules/.bin/jest"
+    "lint": "eslint .",
+    "pretest": "npm run lint",
```
smusali, line 8: @darinspivey, why don't you just put `eslint .` instead of `npm run lint` since you have already removed that part from `.circleci/config.yml` as well?

darinspivey: It's just to reuse the alias.  That way we're not duplicating commands, and `npm run lint` can work on its own.

darinspivey:
> Overall, looks good considering the most of these changes are the linting changes - since you have already touched and  linted all of those files, can you make sure to replace var with let and/or const depending on how it gets used? The next version is gonna get compiled / built using Node v8.3.0 in which there is full support for let.

There is plans to refactor later, and I don't think we should mix efforts, ie. don't do code changes on top of linting changes.  We want the code changes to stand on their own commit, and that can be part of the next code refactor for this (which, honestly is up in the air at this point since there's talk about sunsetting the node version of the agent).

PR #192, opened 2020-08-26
test: Add happy-path integration test

darinspivey:
This test tails a log and makes sure the expected payload is sent
to logDNA.  Other tests were not easily added due to the fact they
are specific to certain OS's and things like k8s.

Semver: patch
Ref: LOG-7033

### DIFF
--- a/index.js
+++ b/index.js
@@ -11,7 +11,7 @@ const properties = require('properties')
 const request = require('request')
 
 // Internal Modules
-const connectionManager = require('./lib/connection-manager')
+const connectionManager = require('./lib/connection-manager.js')
 const k8s = require('./lib/k8s')
 const pkg = require('./package.json')
 const utils = require('./lib/utils')
@@ -21,7 +21,7 @@ const HOSTNAME_IP_REGEX = /[^0-9a-zA-Z\-.]/g
 const HOSTNAME_PATH = '/etc/logdna-hostname'
 
 // Variables
-var config = require('./lib/config')
+let config = require('./lib/config.js')
 var processed
 
 const commander = new Command()
@@ -235,8 +235,15 @@ function loadConfig(program) {
       }
 
       // merge into single var after all potential saveConfigs finished
-      config = {...config, ...parsedConfig}
-      config.tags = process.env.LOGDNA_TAGS || config.tags
+      config = {
+        ...config
+      , ...parsedConfig
+      , tags: process.env.LOGDNA_TAGS || config.tags
+      , flushTimer: null
+      , healthcheckTimer: null
+      , rescanTimer: null
+      }
+
       if (process.env.LOGDNA_PLATFORM) {
         config.platform = process.env.LOGDNA_PLATFORM
         config.tags = config.tags ? `${config.tags},${config.platform}` : config.platform
--- a/lib/connection-manager.js
+++ b/lib/connection-manager.js
@@ -18,8 +18,13 @@ const winUtils = require('./windows-utilities')
 // Variables
 var firstrun
 
+module.exports = {
+  connectLogServer
+}
+
 // Send POST Request to HealthCheck Endpoint
-const checkHealth = (config) => {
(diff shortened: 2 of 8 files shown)


lib/connection-manager.js
```diff
@@ -81,15 +86,15 @@ const checkHealth = (config) => {
...
+      config.healthcheckTimer = setTimeout(() => {
         return checkHealth(config)
       }, config.HEALTHCHECK_INTERVAL) // Every 30 Minutes
     })
   }
 }
 
 // Initiate Connection
-const connectLogServer = (config) => {
+function connectLogServer(config) {
```
mdeltito, line 97: it would be helpful if this (and `checkHealth`) accepted a callback so we didn't need the sleep in tests, and just in general.

out of scope here probably just something that stood out.

index.js
```diff
@@ -21,7 +21,7 @@ const HOSTNAME_IP_REGEX = /[^0-9a-zA-Z\-.]/g
 const HOSTNAME_PATH = '/etc/logdna-hostname'
 
 // Variables
-var config = require('./lib/config')
+let config = require('./lib/config')
```
mdeltito, line 24: 'config.js' since we're updating this line

mdeltito (dismissed):
LGTM, comments are not blockers

lib/connection-manager.js
darinspivey: True.  I can addd a `TODO`.  The code does so much with different platforms that I'm at a loss with how to add more tests without doing the planned `major` release--but that will require solving some issues with our test approach on different OS's.

index.js
darinspivey: sure thing

lib/tailreadstream/tailreadstream.js
```diff
@@ -64,9 +64,10 @@ class TailReadStream extends Readable {
   get timeout() { return this._timeout }
   set timeout(timeout) { this._timeout = timeout }
 
-  destroy() {
+  _destroy() {
```
smusali, line 67: you renamed it but where does it get called? have you changed it there?

index.js
```diff
@@ -235,8 +235,15 @@ function loadConfig(program) {
...
       // merge into single var after all potential saveConfigs finished
-      config = {...config, ...parsedConfig}
-      config.tags = process.env.LOGDNA_TAGS || config.tags
+      config = {
+        ...config
+      , ...parsedConfig
+      , tags: process.env.LOGDNA_TAGS || config.tags
+      , flushTimer: null
+      , healthcheckTimer: null
+      , rescanTimer: null
```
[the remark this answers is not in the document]
darinspivey: We need to store them somewhere so that they can be canceled.  As a general style, objects need to be pre-defined with all properties to avoid map transitions.  Plus, adding these things to `config` on-the-fly gets them lost in code.  So, we add them here 1) to be clear that they will be defined and 2) so that we can cancel them in shutdown.

lib/tailreadstream/tailreadstream.js
darinspivey: This class extends `Readable`, so it's a stream, and this is a mis-implementation of node streams.  The nodejs docs state that `destroy` should NOT be overridden, but it will call `_destroy` automatically if you have custom operations to be done there.  

https://nodejs.org/dist/latest-v10.x/docs/api/stream.html#stream_readable_destroy_error

```
Implementors should not override this method, but instead implement readable._destroy().
```

darinspivey: The integration test also shows that it works :]. Otherwise, the process would not shut down cleanly.

smusali: @darinspivey, is this what can potentially cause the memory leak?

darinspivey: not sure, but it aint good :)

darinspivey: I suppose it could because it's bypassing what the node internals do for `destroy`.

PR #195, opened 2020-09-11
feat: Use the node client for logging

darinspivey:
The node client should be responsible for sending log
entries to the ingester.  Doing this removes lots of
duplicated (and inconsistent) connection management from
the agent.

Semver: minor
Ref: LOG-7104


index.js
```diff
@@ -227,27 +227,31 @@ function loadConfig(program) {
...
       }
 
+      return getos(cb)
+    }
+  , (distro, cb) => {
+      let distroName = `${(distro.dist || distro.os).replace(/Linux| /g, '')}`
+      if (distro.release) {
+        distroName += `/${distro.release}`
+      }
+      config.UserAgent = `${pkg.name}/${pkg.version} (${distroName})`
```
darinspivey, line 241: This value is out of hand, we need to reel it in.  I'm fine with multiple values, but we should simplify and just comma-delimit them rather than try to use the parens.  Ultimately, this will play nicer with the node client, which will put this value in parens.

lib/connection-manager.js
```diff
@@ -1,125 +1,90 @@
...
-
-// Variables
-var firstrun
+const client = require('./logger-client.js')
+const fileUtils = require('./file-utilities.js')
+const utils = require('./utils.js')
 
 module.exports = {
-  connectLogServer
+  checkHealth
```
darinspivey, line 14: Probably don't need to export this because it's coming out soon anyway.  Normally, I'd export it and test it directly to make sure it's switching the log host, but that test will go in the client eventually.

index.js
mdeltito: agreed, this is going to end up gnarly - something like 

```logger-node/x.x.x (logdna-agent/x.x.x (ubuntu/x.x.x) k8s)```

+1 for just simple comma-delimited `package/version` in the future. that would give us all the same context we need in reporting/tooling. will require support on the ingestion side.

File: lib/connection-manager.js [hunk 3]
```diff
@@ -1,125 +1,90 @@
...
-      , qs: {
-          timestamp: Date.now()
-        }
-      }, (error, response, body) => {
-        if (error) return reqCallback(error.code)
-        if (response && response.statusCode >= 500) return reqCallback('INTERVAL_SERVER_ERROR')
-        return reqCallback(null, {...JSON.parse(body || '{}'), statusCode: response.statusCode})
+    , body: JSON.stringify({
+        e: 's'
+      , m: process.memoryUsage()
```
mdeltito, line 41: do we need to continue sending this with the removal of restarts? I'm not sure of the cost of `process.memoryUsage()`

smusali: yea, it seems like we don't do anything with that info - it was needed for `autoRestart` but now we are getting rid of that, no?

lib/connection-manager.js
smusali: and what is preventing us removing the `checkHealth` from the code base, @darinspivey?

lib/utils.js
```diff
@@ -110,27 +108,6 @@ const processOption = (options, config, hasWhitespace) => {
   }
 }
 
-// Restart LogDNA Agent
```
smusali, line 113: @darinspivey, can you remove the references to these `restartSelf` and `updateSelf` methods from `README.md` and some other places as well?

lib/config.js
```diff
@@ -6,34 +6,25 @@ const path = require('path')
...
 , FILEQUEUE_PATH: os.platform() !== 'win32'
     ? (process.env.FILEQUEUE_PATH || '/tmp')
     : path.join(process.env.ALLUSERSPROFILE, (process.env.FILEQUEUE_PATH || '/tmp'))
-, FLUSH_INTERVAL: process.env.FLUSH_INTERVAL || 250 // 250 millisec
-, FLUSH_LIMIT: process.env.FLUSH_LIMIT || 5000 // 5000 lines
-, GZIP_COMPRESS_LEVEL: process.env.GZIP_COMPRESS_LEVEL || 2
+, FLUSH_INTERVAL: process.env.FLUSH_INTERVAL
+, FLUSH_LIMIT: process.env.FLUSH_LIMIT
 , HEALTHCHECK_INTERVAL: process.env.HEALTHCHECK_INTERVAL || 1800000 // 30 minutes
-, HTTP_KEEPALIVE: process.env.HTTP_KEEPALIVE || 60000 // 1 min
```
smusali, line 24: @darinspivey, what are replacing these `HTTP_KEEPALIVE`, `MAX_SOCKETS`, `SOCKET_KEEPALIVE` in the node client?

smusali (changes requested):
Overall, looks good but having few comments - mainly, wondering about why not completely getting rid of `healthcheck` calls?

lib/connection-manager.js
darinspivey: I'll remove!

lib/utils.js
darinspivey: yes, good catch

lib/config.js
darinspivey: I'll have a look to see what can be passed through to the client.

darinspivey: These aren't supported in the client, and honestly, I think we should go with the defaults for those things.  `agentkeepalive` might be switched out in the future, and these are specific to it.  This being a `major`, I think this is a safe move.

lib/utils.js
darinspivey: Removed one line about auto-upgrading.

index.js
darinspivey: Fixed.

lib/connection-manager.js
darinspivey: Removed entirely!

README.md
```diff
@@ -70,10 +72,13 @@ key = <YOUR LOGDNA INGESTION KEY>
...
+   files. By default this option is set to monitor `.log` and extensionless files under `/var/log/`.
+* `exclude`: sets files to exclude that would otherwise match what's set in `logdir`. Multiple paths can be specified, separated by `,`.
+   Supports glob patterns + specific files
 * `exclude_regex`: filters out any log lines matching this pattern in any file. Should not include leading or trailing `/`.
-* `key`: your LogDNA Ingestion Key. You can obtain one by creating an account at [LogDNA](https://logdna.com/). Once logged in, click on the Gear icon, then Account Profile to find your key.
+* `key`: your LogDNA Ingestion Key. You can obtain one by creating an account at [LogDNA](https://logdna.com/). Once logged in, click on the Gear icon,
+   then Account Profile to find your key.
 * `tags`: tags can be used e.g. to separate data from production, staging, or autoscaling use cases
 * `hostname`: set this to override the os hostname
 * `autoupdate`: whether the agent should update itself when new versions are available on the public repo (default is `1`, set to `0` to disable)
```
smusali, line 84: @darinspivey, what about this?

lib/config.js
smusali: Sounds good!

README.md
```diff
@@ -70,10 +72,13 @@ key = <YOUR LOGDNA INGESTION KEY>
 ```
 
 #### Configuration File Options
-* `logdir`: sets the paths that the agent will monitor for new files. Multiple paths can be specified, separated by `,`. Supports glob patterns + specific files. By default this option is set to monitor `.log` and extensionless files under `/var/log/`.
-* `exclude`: sets files to exclude that would otherwise match what's set in `logdir`. Multiple paths can be specified, separated by `,`. Supports glob patterns + specific files
+* `logdir`: sets the paths that the agent will monitor for new files. Multiple paths can be specified, separated by `,`. Supports glob patterns + specific
```
dm36, line 75: is there an order we have for these options- or an order that would make most sense?

README.md
```diff
@@ -103,7 +107,8 @@ sudo apt-get update
 
 ## How it Works
 
-The LogDNA agent authenticates using your [LogDNA Ingestion Key](https://app.logdna.com/manage/profile) and opens a secure web socket to LogDNA's ingestion servers. It then 'tails' for new log data, as well as watches for new files added to your specific logging directories.
+The LogDNA agent authenticates using your [LogDNA Ingestion Key](https://app.logdna.com/manage/profile) and opens a secure web socket to LogDNA's
+ingestion servers. It then 'tails' for new log data, as well as watches for new files added to your specific logging directories.
```
dm36, line 112: What was changed here?

README.md
```diff
@@ -56,7 +56,9 @@ Note that when using glob patterns with `index.js`, you must enclose the pattern
 
 ### Configuration File
 
-Normally a config file is automatically generated (e.g. when you set a key using `index.js -k`) and updated (e.g. when you add a directory using `index.js -d`) but you can create your own config file `/etc/logdna.conf` on Linux and `C:\ProgramData\logdna\logdna.conf` on Windows and save your settings there:
+Normally a config file is automatically generated (e.g. when you set a key using `index.js -k`) and updated (e.g. when you add a directory
+using `index.js -d`) but you can create your own config file `/etc/logdna.conf` on Linux and `C:\ProgramData\logdna\logdna.conf` on Windows and
+save your settings there:
```
dm36, line 61: Are you adding newlines so that it's all not on a single line?

dm36: Are you arbitrarily cutting off lines based off of length or using a ruler- I know they have those in Sublime/VScode

index.js
```diff
@@ -288,7 +288,7 @@ function loadConfig(program) {
     }
 
     debug('connecting to log server')
-    connectionManager.connectLogServer(config)
+    setupLoggerThenStart(config)
```
dm36, line 291: maybe InitializeLogger? Am thinking about naming for this function

README.md
darinspivey: I usually order them in the order that they are in the code, but there's really no rule.  Alphabetizing is another option.  We will be re-writing all of these docs in another PR.

lib/file-utilities.js
```diff
@@ -117,10 +117,12 @@ function streamFiles(config, logfiles, callback) {
...
       tail = spawn('tail', ['-n0', '-F', file])
       tail.stdout.on('data', (data) => {
-        data = data.toString().trim().split('\n')
-        for (i = 0; i < data.length; i++) {
-          data[i] = {t: Date.now(), l: data[i], f: file, label: labels}
-          linebuffer.addMessage(data[i])
+        const lines = data.toString().trim().split('\n')
+        for (i = 0; i < lines.length; i++) {
+          client.log.agentLog({
+            line: lines[i]
```
dm36, line 123: the node-client will auto-add the timestamp

README.md
darinspivey: *shakes fist* :)

README.md
darinspivey: Yeah, this was all arbitrarily just cutting lines longer than 160 for readability.  We will eventually re-write all of this, but that shouldn't be until we have a solidified product (perhaps after, or right before, we split this repo).

index.js
darinspivey: But this is the entrypoint for also starting the scanning of the logs, so that's important to have in the name.  It's also worth nothing that, although this is a small function, it's necessary because it helps testing.  It cuts out all of the cruft in index.js that we don't need or want for every test.

darinspivey: If you like the word `initialize` better, I'm happy to change it.

lib/file-utilities.js
darinspivey: yep.  There were many avenues to pass `timestamp`, so I'm letting the client do it.

lib/file-utilities.js
```diff
@@ -243,7 +243,7 @@ function streamAllLogs(config, callback) {
       for (i = 0; i < data.length; i++) {
```
mdeltito, line 243: not blocking on this but it looks like we dont need the index anywhere, so could simplify with `for..of` here and in `streamFiles`.

mdeltito (dismissed):
this looks great! net removal of 400+ lines looks even better

lib/file-utilities.js
darinspivey: Ah, that's part of journald.  I think I'd rather not mess up the scope of this work with those changes because we're deleting that code coming soon.  Lemme know if you feel strongly otherwise.

index.js
darinspivey: actually, I'll get consistent with this and call it `createLoggerThenStart` cause that's what it does :]. That is the closest to the actual function names too.

mdeltito (approved):

lib/config.js
```diff
@@ -6,34 +6,25 @@ const path = require('path')
 
 module.exports = {
   AWS_INSTANCE_CHECK_URL: 'http://169.254.169.254/latest/dynamic/instance-identity/document/'
-, BUF_LIMIT: process.env.BUF_LIMIT || 10000 // 10000 lines
-, COMPRESS: isNaN(process.env.COMPRESS) ? true : +process.env.COMPRESS
+, COMPRESS: isNaN(process.env.COMPRESS) || process.env.COMPRESS === '1'
```
smusali, line 9: is there any example for the possible values of `process.env.COMPRESS`? If not, can you add?

index.js
```diff
@@ -227,28 +227,28 @@ function loadConfig(program) {
...
+      if (distro.release) {
+        distroName += `/${distro.release}`
+      }
+      config.UserAgent = `${pkg.name}/${pkg.version} ${distroName}`
+
       if (process.env.LOGDNA_PLATFORM) {
-        config.platform = process.env.LOGDNA_PLATFORM
-        config.tags = config.tags ? `${config.tags},${config.platform}` : config.platform
-        if (config.platform.indexOf('k8s') === 0) {
+        const platform = process.env.LOGDNA_PLATFORM
```
smusali, line 244: `config.platform` was being mainly used for `k8s` but since we are gonna remove it, I think, no need for `config.platform` but need to double check where else it gets used

index.js
```diff
@@ -288,7 +288,7 @@ function loadConfig(program) {
     }
 
     debug('connecting to log server')
-    connectionManager.connectLogServer(config)
+    createLoggerThenStart(config)
```
smusali, line 291: can we just name it as `start`? it seems like `createLoggerThenStart` indicates that there is another `...Start` method in which no logger creation.

I mean, we already know that it won't start anything without creating logger - so, `start(config)` would be better, imo

lib/utils.js
```diff
@@ -190,38 +167,11 @@ const unsetConfig = (options, config) => {
...
-    , '-y'
-    , '--allowunofficial'
-    ], {
-      detached: true
-    })
-  }
-}
-
 exports.appender = appender
 exports.buildUrl = buildUrl
```
smusali, line 171: I think no need for this method anymore since we are using node client

smusali (commented):
few comments but overall, looks good

lib/config.js
darinspivey: so this is why we need the `config` map.  It's a boolean.  Where do you want me to add the possible values?

index.js
darinspivey: Thanks.  I can take it out on the next PR which will be k8s specific.  Lemme know what you find.

index.js
darinspivey: I'll buy that.  Naming is hard.

lib/utils.js
darinspivey: The node client expects a full URL, and since we have it broken up by host, port in the env, we have to build the string.  I'd rather leave it for now, but I'd like to just make this the full URL eventually.  The host and port were needed for the healthcheck now, and that's gone.  Should we tweak it now as part of this major?  maybe....

lib/config.js
smusali: I was thinking about adding examples to README.md if there is specific section for env variables but it seems there is not + this config part is gonna get rewritten, right?

lib/utils.js
smusali: It will deprecate some env variables, right? The env variables which are frequently used by customers - we can easily include that in one of our upcoming major releases but we should clearly state what's been changed - in very noticeable way

darinspivey: Lemme see about cleaning that up now.  I also forgot to take the healthcheck constant out.

index.js
darinspivey: done

lib/utils.js
darinspivey: I ripped off the bandaid and make it a full URL.  This removes even more code!

lib/config.js
```diff
@@ -6,36 +6,20 @@ const path = require('path')
...
-, LOGDNA_LOGENDPOINT: process.env.LDLOGPATH || process.env.INGESTION_ENDPOINT || '/logs/agent'
-, LOGDNA_LOGHOST: process.env.LDLOGHOST || process.env.INGESTION_HOST || 'logs.logdna.com'
-, LOGDNA_LOGPORT: process.env.LDLOGPORT || process.env.INGESTION_PORT || 443
-, LOGDNA_LOGSSL: isNaN(process.env.LDLOGSSL) ? true : +process.env.LDLOGSSL
-, LOGDNA_RECONNECT: true
-, MAX_SOCKETS: process.env.MAX_SOCKETS || 20
+, FLUSH_INTERVAL: process.env.FLUSH_INTERVAL
+, FLUSH_LIMIT: process.env.FLUSH_LIMIT
+, PROXY: process.env.HTTPS_PROXY || process.env.HTTP_PROXY
+, LOGDNA_INGESTION_URL: process.env.LOGDNA_INGESTION_URL || 'https://logs.logdna.com/logs/agent'
```
smusali, line 19: Sorry for nitpicking on naming but what about keeping the config variable / parameter name consistent across the integrations? Since we are removing all `LOGDNA_*` parameters, why to call this `LOGDNA_INGESTION_URL`?

On `@logdna/logger-node`, it's `url`; on `logspout` and lambdas, it's `LOGDNA_URL` since they are also environment variables there - as you mentioned, naming is hard but I'd suggest to call it `LOGDNA_URL`.

lib/utils.js
```diff
@@ -68,6 +55,11 @@ const log = (message, level) => {
   const time = `${padStart(dateObject.getHours(), 2)}:${padStart(dateObject.getMinutes(), 2)}:${padStart(dateObject.getSeconds(), 2)}`
   const tzOffset = dateObject.getTimezoneOffset() * -100 / 60
   const timezone = `${(tzOffset > 0 ? '+' : '')}${padStart(tzOffset, 4)}`
+  // Properly handle Error objects.  The client may have additional meta information to log
+  if (typeof message === 'object' && message.constructor && message.constructor.name === 'Error') {
```
smusali, line 59: @darinspivey, can you show me as an example where this kind of thing can happen in the newer changes? I mean, where is `message` gonna be an `object`?

smusali: Just as an extra context: these `console` statements generates the logs which are getting into `/var/log/logdna-agent.log` from which they get streamed into and parsed on `LogDNA` where we parse `datetime`, `timezone`, and `level` info

smusali (changes requested):
By the way, `glob`-specific testing error happened

lib/utils.js
darinspivey: While testing, I got a nock error, and all that came out was "there was a problem" from the client.  [Here is the place](https://github.com/logdna/logdna-agent/blob/e7ae9b24b7c0f0f9acb996402058d8e20b1f92e1/lib/logger-client.js#L32) where this change will benefit us.  It will now look like this:

```
2020-09-17 12:50:18 -0400 [error] { Error: An error occured while sending logs to the cloud.
    at process.nextTick (/Users/dspivey/github/logdna-agent/node_modules/@logdna/logger/lib/logger.js:719:25)
    at process._tickCallback (internal/process/next_tick.js:61:11)
  meta:
   { actual:
      'Nock: No match for request {\n  "method": "POST",\n  "url": "https://logs.logdna.com/logs/agent?now=1600361418306&hostname=MyHostMachine&mac=... [snip]\\"}]}"\n}',
     code: 'ERR_NOCK_NO_MATCH',
     firstLine: 'Here is my line1',
     lastLine: 'Here is my line 2',
     retrying: false,
     attempts: 1 } }
```

lib/config.js
darinspivey: ok, that's fine

lib/config.js
darinspivey: Yeah, we're gonna re-write all of this.  And the docs are severely lacking now, which is why I hope the config package will clear lots of this up (because it generates docs)

darinspivey:
> By the way, `glob`-specific testing error happened

That glob error is intentional for testing.  The error was the tap thresholds are a bit off in CI versus local.  Whatever.

lib/utils.js
darinspivey: BTW, the changes to the env vars, I'd assumed would all go in the CHANGELOG

lib/utils.js
smusali: I see @darinspivey but if possible why don't we change how we log the error [here](https://github.com/logdna/logdna-agent/blob/e7ae9b24b7c0f0f9acb996402058d8e20b1f92e1/lib/logger-client.js#L32) instead of touching this `log` method since only newer change `(logger-client)` addition created necessity to add this handling but I don't think it's gonna be used anywhere else?

mdeltito: afaict this isn't necessary other than to use `console.error` instead, is that right? without this, the full error would still be passed to `console.log` now with the other changes

smusali: No extra concerns here - not gonna block this and it's easy to update whenever needed and/or requested in the future

smusali (approved):
lgtm

mdeltito (approved):

lib/utils.js
darinspivey: Sorry, @mdeltito , forgot to answer you.  Yes, the big thing here was using `console.error`.  With `console.log`, the full error was being truncated for some reason.  We *really* need to use pino, but our pino config is not public, however that shouldn't stop us from doing it.

ankurgarg21:
HI There, 

After LogDNA agent upgraded, it is not working, can you check and advise ?

[vagrant@localhost ~]$ rpm -qa | grep logdna-agent
logdna-agent-2.0.0-1.x86_64
[vagrant@localhost ~]$ cat /etc/redhat-release 
CentOS Linux release 7.8.2003 (Core)
[vagrant@localhost ~]$

[vagrant@localhost ~]$ sudo logdna-agent 
2020-10-09 09:18:41 0000 [info] logdna-agent/2.0.0 started on abcd.com (0.0.0.0)
2020-10-09 09:18:41 0000 [info] LogDNA URL: https://logs.logdna.com/logs/agent
2020-10-09 09:18:41 0000 [info] streaming /var/log: 48 file(s)
2020-10-09 09:18:41 0000 [info] streaming /data/jenkins/logs: 5 file(s)
2020-10-09 09:18:44 0000 [error] { Error: An error occured while sending logs to the cloud.
    at process.nextTick (/usr/bin/node_modules/@logdna/logger/lib/logger.js:719:25)
    at _combinedTickCallback (internal/process/next_tick.js:131:7)
    at process._tickCallback (internal/process/next_tick.js:180:9)
  meta: 
   { actual: 'Request failed with status code 403',
     code: 403,
     firstLine: '2020-10-09 09:18:42 0000 [error] { Error: An error occured while sending logs to the cloud.',
     lastLine: '     attempts: 1 } }',
     retrying: false,
     attempts: 1 } }
2020-10-09 09:18:44 0000 [error] { Error: An error occured while sending logs to the cloud.
    at process.nextTick (/usr/bin/node_modules/@logdna/logger/lib/logger.js:719:25)
    at _combinedTickCallback (internal/process/next_tick.js:131:7)
    at process._tickCallback (internal/process/next_tick.js:180:9)
  meta: 
   { actual: 'Request failed with status code 403',
     code: 403,
     firstLine: '2020-10-09 09:18:43 0000 [error] { Error: An error occured while sending logs to the cloud.',
     lastLine: '     attempts: 1 } }',
     retrying: false,
     attempts: 1 } }
^C2020-10-09 09:18:44 0000 [info] got SIGINT signal, shutting down...

PR #196, opened 2020-09-17
fix: remove kubernetes support

darinspivey:
Most kubernetes processing is now done with our Rust agent.
Removing that here allows us to stop building a docker
container, which currently is preventing us from being
able to properly version this repo due to the tagging and
naming of the docker image.  Customers needing k8s support
should use the rust agent.

Semver: major
Ref: LOG-7324


README.md
```diff
@@ -117,21 +117,6 @@ logdna register <email>
 # now paste the Ingestion Key into the kubectl commands above
```
smusali, line 117: need to remove this line too

README.md
```diff
@@ -117,21 +117,6 @@ logdna register <email>
...
 ```
 
-## Kubernetes Logging
```
smusali, line 120: @darinspivey, I'm still not 100% sure whether it's good idea to remove this whole block since we just stop building new Docker images and there are still `yamls` inside this repo. Plus, how consistent is this with our other docs for `k8s`? Should we have separate PR for this?

README.md
```diff
@@ -117,21 +117,6 @@ logdna register <email>
...
-kubectl create secret generic logdna-agent-key --from-literal=logdna-agent-key=<YOUR LOGDNA INGESTION KEY>
-kubectl create -f https://raw.githubusercontent.com/logdna/logdna-agent/master/logdna-agent-ds.yaml
-```
-
-This automatically installs a logdna-agent pod into each node in your cluster and ships stdout/stderr from all containers, both application
-logs and node logs. Note: by default, the agent pod will collect logs from all namespaces on each node, including `kube-system`. View your logs
-at https://app.logdna.com. See [YAML file](https://raw.githubusercontent.com/logdna/logdna-agent/master/logdna-agent-ds.yaml) for additional options
-such as `LOGDNA_TAGS`.
-
 ### Upgrading to LogDNA Agent 2.0 for Kubernetes
```
smusali, line 120: if we are removing that block, why are we keeping this one? I'm wondering...

index.js
```diff
@@ -240,15 +239,6 @@ function loadConfig(program) {
       }
       config.UserAgent = `${pkg.name}/${pkg.version} ${distroName}`
 
-      if (process.env.LOGDNA_PLATFORM) {
```
smusali, line 243: @darinspivey, people use `LOGDNA_PLATFORM` for some other platforms as well based on what I have seen in audits - it's not just for `k8s`

index.js
```diff
@@ -281,13 +271,11 @@ function loadConfig(program) {
     }
 
     utils.log(`${config.package} started on ${config.hostname} (${config.ip})`)
-    if (config.platform && config.platform.startsWith('k8s')) { k8s.init() }
     if (config.userAgent) {
       config.DEFAULT_REQ_HEADERS['user-agent'] = config.userAgent
       config.DEFAULT_REQ_HEADERS_GZIP['user-agent'] = config.userAgent
     }
 
-    debug('connecting to log server')
```
smusali, line 290: I'd say let's keep this debug statement here - unrelated to `K8s` support and if we remove this from here, we need to add something similar to `start` functionality showing that the execution reached this point to make connection with the server

lib/utils.js
```diff
@@ -103,22 +113,21 @@ const processOption = (options, config, hasWhitespace) => {
...
+function saveConfig(config, configPath, callback) {
   return properties.stringify(config, {
     path: configPath
   }, callback)
 }
 
 // Custom JSON Stringifier
-const stringify = (obj) => {
-
+// TODO(darinspivey) Why do we need this?  We shouldn't roll our own stringifier.
```
smusali, line 123: @darinspivey, the intention was to properly show the configuration in the command line - well-aligned, clean, etc. - we will need to wipe this out, most probably, in the config work.

smusali (changes requested):
Some comments to be addressed...

README.md
darinspivey: I think it's *really* confusing to leave stuff in (although I'm fine with keeping the yamls).  If this repo no longer does k8s for v1, we shouldn't keep saying that it does.  I think we're covered by the `major` change for this.

README.md
darinspivey: Well, this is for v2 which is still fully supported.  Man, this is confusing.

index.js
darinspivey: ok.  I'll just remove the k8s logic then.

index.js
darinspivey: sure.  I just removed it because it's not accurate.  "connect" makes it seem persistent, but it's just a POST endpoint.  I can keep it, but re-word it.

lib/utils.js
darinspivey: I get it, but that goes back to programming for human eyes when we should first write code for computers.  This kind of custom logic tends to stick around and become a burden.  We'll look closer at it in the future in case I'm missing something.  I didn't look hard at it, it just feels weird to re-invent that wheel.

README.md
darinspivey: Not a bad idea, @mdeltito , but we can't just link people to the repo, right?  They'd be interested in the docker image, not the code base, amirite?

mdeltito: well I mean a link to the readme over there, which seems to have all the same info for the rust kube agent (installing on k8s via yaml files which specify the image)

smusali: I agree to have `./docs` folder - it's time to have that - they may wanna see the code - so, providing links to the images and codebase and this explanation should be enough - thanks, @mdeltito, for great idea!

README.md
smusali: `./docs` should be enough to cover that side of the story - so, nothing to worry

README.md
darinspivey: done

mdeltito (dismissed):

docs/kubernetes.md
```diff
@@ -0,0 +1,38 @@
...
+you may use our LogDNA command line interface to register:
+
+```
+brew cask install logdna-cli
+logdna register <email>
+# now paste the Ingestion Key into the kubectl commands above
+```
+
+We created this integration mainly based on customer feedback and that
+[logging with Kubernetes should not be this painful](https://blog.logdna.com/2017/03/14/logging-with-kubernetes-should-not-be-this-painful/).
```
smusali, line 38: @darinspivey, the link is broken here - the correct link is [this](https://logdna.com/blog/logging-with-kubernetes-should-not-be-this-painful/) but it is missing some images there + it contains the `Node-based Agent` instructions for `K8s`

I'd suggest to remove ` and ...` from the sentence

smusali (commented):
small comment

docs/kubernetes.md
darinspivey: I'm just gonna remove the whole last sentence.  It doesn't offer value, IMO

smusali (approved):
lgtm

mdeltito (approved):

PR #197, opened 2020-09-25
fix: General fixes prior to releasing v1, 2.0

darinspivey:
* Tags were not being passed to the client due to an
  incorrect variable name.
* A log entry had a blank `config.package`.  Removed.
* The `hostname` regex was too strict.  Added dots and dashes.

Semver: patch
Ref: LOG-7387

### DIFF
--- a/lib/logger-client.js
+++ b/lib/logger-client.js
@@ -17,7 +17,7 @@ function createLoggerClient(config) {
   logger = createLogger(config.key, {
     payloadStructure: 'agent'
   , url: config.LOGDNA_URL
-  , hostname: config.hostname.replace(/[^a-zA-Z0-9]/g, '') // Satisfy hostname regex in the client
+  , hostname: config.hostname.replace(/[^a-zA-Z0-9.-]/g, '') // Satisfy hostname regex in the client
   , mac: config.mac
   , ip: config.ip
   , tags: config.tags
--- a/index.js
+++ b/index.js
@@ -225,7 +225,7 @@ function loadConfig(program) {
       config = {
         ...config
       , ...parsedConfig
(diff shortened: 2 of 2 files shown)


lib/logger-client.js
```diff
@@ -17,7 +17,7 @@ function createLoggerClient(config) {
   logger = createLogger(config.key, {
     payloadStructure: 'agent'
   , url: config.LOGDNA_URL
-  , hostname: config.hostname.replace(/[^a-zA-Z0-9]/g, '') // Satisfy hostname regex in the client
+  , hostname: config.hostname.replace(/[^a-zA-Z0-9.-]/g, '') // Satisfy hostname regex in the client
```
ydshah2, line 20: Are we not putting any constraints on use of . and -? like start or end with a dash or not?

mdeltito: should we just use the same regex as the client?
https://github.com/logdna/logger-node/blob/3bf0ecfb6cc417f97f7f974b11ac2cf7107810eb/lib/constants.js#L14

darinspivey: TBH, I don't think `hostname` should have any constraints.  We don't want to have the client throw an error if it's too strict.  We're gonna have to look into the ingester to see if there was a reason why we're only accepting a certain format. For now, this change will still appease the `logger` while being a bit more flexible.

mdeltito: works for me, and I agree that validating the hostname in general on the client side doesn't gain us much. the length is the only restriction on the ingestion side but it just gets trimmed.

mdeltito (approved):

matt-march (approved):

PR #198, opened 2020-09-25
feat: Use eslint-config-logdna from NPM

darinspivey:
Our linting rules are now public.  Start using it.

Semver: patch
Ref: LOG-7396


mdeltito (approved):
🎉

matt-march (approved):

PR #199, opened 2020-09-29
config: upgrade node version to v12.16.2

smusali:
After fully testing on supported OS versions, we have decided to upgrade the node version to v12.16.2 for MacOSX, Windows, and Debian Systems. This Pull Request contains some small improvements and fixes
too.

Ref: LOG-7309
Semver: major

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,9 +1,6 @@
-# logdna-agent
+# LogDNA Agent
 
-[![Build Status](https://travis-ci.org/logdna/logdna-agent.svg?branch=master)](https://travis-ci.org/logdna/logdna-agent)
-[![Build status](https://ci.appveyor.com/api/projects/status/mk5rb0uk6xkjxhk2/branch/master?svg=true)](https://ci.appveyor.com/project/mikehu/logdna-agent/branch/master)
-
-LogDNA's collector agent which streams log files to your LogDNA account. LogDNA is a hosted, cloud logging service.
+LogDNA Agent streams from log files to your LogDNA account. Works with Linux, Windows, and macOS Servers.
 
 ## Getting Started
 
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "logdna-agent",
   "version": "1.6.5",
-  "description": "LogDNA Collector Agent",
+  "description": "LogDNA Agent streams from log files to your LogDNA account. Works with Linux, Windows, and macOS Servers",
   "main": "index.js",
   "scripts": {
     "lint": "eslint .",
@@ -36,11 +36,9 @@
   },
   "dependencies": {
     "@logdna/logger": "^1.3.2",
-    "agentkeepalive": "^4.1.0",
     "async": "^3.2.0",
     "commander": "^4.1.1",
     "debug": "^4.1.1",
-    "file-queue": "^0.3.0",
     "getos": "^3.1.4",
     "glob": "^7.1.6",
     "is-administrator": "^1.0.1",
--- a/tools/scripts/darwin.sh
+++ b/tools/scripts/darwin.sh
@@ -5,12 +5,12 @@
 ARCH=x64
 INPUT_TYPE=dir
 LICENSE=MIT
-NODE_VERSION=8.3.0 # Will upgrade after 1.6.5
+NODE_VERSION=12.16.2
 OSXPKG_IDENTIFIER_PREFIX=com.logdna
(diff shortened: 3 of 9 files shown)


package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "logdna-agent",
   "version": "1.6.5",
-  "description": "LogDNA Collector Agent",
+  "description": "LogDNA Agent which streams from log files to your LogDNA account. Works with Linux, Windows, and macOS Servers",
```
darinspivey, line 4: I'd remove `which` from this.  `LogDNA Agent streams...`

darinspivey (changes requested):

package.json
smusali: @darinspivey, done!

tools/files/darwin/com.logdna.logdna-agent.plist
```diff
@@ -15,8 +15,8 @@
         <key>RunAtLoad</key>
         <true/>
         <key>StandardErrorPath</key>
-        <string>/Library/Logs/logdna-agent.log</string>
+        <string>/var/log/logdna-agent.log</string>
```
mdeltito, line 18: Nothing to do here for now, but for Homebrew best-practices this directory should be within the brew install path. By default that would be `/usr/local/var/log` but users can have a different base directory for their homebrew install. The same is true for the `ProgramArguments` path, that should also take into consideration the hombrew prefix path.

The `nginx` formula takes the approach of generating the plist at install-time for this reason 
https://github.com/denji/homebrew-nginx/blob/834eb3f2f4542d3e3ab5e9d9efd127a9cebc19cd/Formula/nginx-full.rb#L379-L401

Unrelated, line 10 above `com.logdna.logdna-agentd` - is that correct, with the extra `d` at the end?

tools/scripts/darwin.sh
```diff
@@ -58,7 +58,7 @@ sed "s/${OLDSHA256CHECKSUM}/${SHA256CHECKSUM}/" ../tools/files/darwin/logdna-age
 cd ..
 
 # STEP 4: RELEASE
-ghr -draft \
```
mdeltito, line 61: just curious, why is the `draft` step being removed? does this make the `pause` below unnecessary, and/or should we keep the current approval process?

README.md
```diff
@@ -1,6 +1,5 @@
 # logdna-agent
 
-[![Build Status](https://travis-ci.org/logdna/logdna-agent.svg?branch=master)](https://travis-ci.org/logdna/logdna-agent)
 [![Build status](https://ci.appveyor.com/api/projects/status/mk5rb0uk6xkjxhk2/branch/master?svg=true)](https://ci.appveyor.com/project/mikehu/logdna-agent/branch/master)
```
mdeltito, line 3: this is a dead link

smusali: Oh, I thought the other one is CircleCI

tools/scripts/darwin.sh
smusali: `-draft` creates a new `GitHub Draft Release` - I am gonna create a `GItHub Release` manually and since running these scripts manually, I wanna test whether this just attaches the artifacts to already existing release or not

mdeltito: a draft is the same as a "normal" release, just without it being officially published. it's good for testing things like you mentioned safely without publishing a new release, and it can be published later if testing goes well.

that said, no problem from me if you're ok with the process.

smusali: @mdeltito, yea, we will visit that later in the case of automation - right now this is pretty experimental and manual with just some set of automation

tools/files/darwin/com.logdna.logdna-agent.plist
smusali: @mdeltito, decided to return it back - I think, after split, we will have more opportunity to follow the best practices; currently, I have decided not touch and this PR isn't place for touching this segment.

README.md
smusali: Done!

tools/files/darwin/com.logdna.logdna-agent.plist
mdeltito: I agree, and that sounds good. just wanted to mention it while I'm looking at it

mdeltito (approved):

darinspivey (approved):

PR #202, opened 2020-09-30
fix broken link in openshift install

ac427:
Signed-off-by: Ananta Chakravartula <ananta.c@gmail.com>


smusali (approved):
Thanks for the fix, @ac427!

mdeltito (approved):
thanks!

Issue #206, opened 2020-10-09
Logdna agent 2.0 breaks for on-prem installs


darakian:
Hey all,

Just got through an incident where logs were no longer arriving for an "on-prem " install of your service. The issue is something to do with the 2.0 release which we install through your repos. We were seeing logs of the sort

```
[error] { Error: An error occured while sending logs to the cloud.
    at process.nextTick (/usr/bin/node_modules/@logdna/logger/lib/logger.js:719:25)
    at _combinedTickCallback (internal/process/next_tick.js:131:7)
    at process._tickCallback (internal/process/next_tick.js:180:9)
  meta:
   { actual: 'Request failed with status code 403',
     code: 403,
     firstLine: 'Oct  9 17:25:47 ip-172-24-4-240 polkitd[1056]: Registered Authentication Agent for unix-process:27029:844371184 (system bus name :1.14602 [/usr/bin/pkttyagent --notify-fd 6 --fallback], object path /org/freedesktop/PolicyKit1/AuthenticationAgent, locale en_US.UTF-8)',
     lastLine: '2020-10-09 17:25:47 0000 [info] got SIGTERM signal, shutting down...',
     retrying: false,
     attempts: 1 } }
```
with the 2.0 agent. After downgrading back to 1.6.5 we see
```
2020-10-09 17:43:46 0000 [info] logdna-agent/1.6.5 started on some_host (some_ip)
2020-10-09 17:43:46 0000 [info] streaming /var/log/: 49 file(s)
2020-10-09 17:43:49 0000 [info] got SIGTERM signal, shutting down...
2020-10-09 17:43:50 0000 [info] logdna-agent/1.6.5 started on some_host (some_ip)
2020-10-09 17:43:50 0000 [info] streaming /var/log/: 49 file(s)
2020-10-09 17:43:56 0000 [info] got SIGTERM signal, shutting down...
2020-10-09 17:43:56 0000 [info] logdna-agent/1.6.5 started on some_host (some_ip)
2020-10-09 17:43:56 0000 [info] streaming /var/log/: 49 file(s)
```
This seems to be the case across all rhel/debian based hosts. Our logs are sent to a domain of the form `a.b.c.d` which differs from your regular `logs.logdna.com` path. Perhaps that's the issue.

darakian:
Perhaps the removal of `LOGDNA_LOGHOST` here

[...]

PR #208, opened 2020-10-15
fix: Revert environment variable changes

darinspivey:
Previously, we attempted to move to a single URL
for the logger client and remove the many env vars
that previously "built" the URL.  However, doing a `major`
release for this wasn't sufficient enough to not break
customers, so we are putting them back until such time
we can ensure a safer upgrade path for such breaking
changes.

Semver: patch
Fixes: https://github.com/logdna/logdna-agent/issues/206
Ref: LOG-7618

----------------
feat: Retry reading when file is locked
    
This fixes a bug in the Windows version when applications have the
file open and obtain a lock. The solution is simply to wait until 
the process has released the lock and read again.
    
Fixes: https://github.com/logdna/logdna-agent/issues/167
Semver: patch

### DIFF
--- a/.taprc
+++ b/.taprc
@@ -14,7 +14,7 @@ coverage-report:
 - html
 files:
 - test/**/*.js
-statements: 70
+statements: 69
 branches: 57
 functions: 74
-lines: 70
+lines: 69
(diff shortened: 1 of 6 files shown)


smusali (approved):
lgtm

mdeltito (approved):

PR #210, opened 2020-10-28
[Duplicate] Use systemd service name as the app name

smusali:
Considering the lack of activity on [#169](https://github.com/logdna/logdna-agent/pull/169) and the need for rebasing, we needed to have a separate Pull Request for resolving [#168](https://github.com/logdna/logdna-agent/pull/168)

### DIFF
--- a/lib/file-utilities.js
+++ b/lib/file-utilities.js
@@ -223,7 +223,7 @@ function streamAllLogs(config, callback) {
       return {
         t: data.__REALTIME_TIMESTAMP
       , line: data.MESSAGE
-      , f: data.CONTAINER_NAME || data.SYSLOG_IDENTIFIER || 'UNKNOWN_SYSTEMD_APP'
+      , f: data.CONTAINER_NAME || data._SYSTEMD_UNIT || data.SYSLOG_IDENTIFIER || 'UNKNOWN_SYSTEMD_APP'
       , pid: data._PID && parseInt(data._PID)
       , prival: data.PRIORITY && parseInt(data.PRIORITY)
       , containerid: data.CONTAINER_ID_FULL


darinspivey (changes requested):
Nit, but can you make your commit body 72 characters or less, please?

lib/file-utilities.js
```diff
@@ -223,7 +223,7 @@ function streamAllLogs(config, callback) {
       return {
         t: data.__REALTIME_TIMESTAMP
       , line: data.MESSAGE
-      , f: data.CONTAINER_NAME || data.SYSLOG_IDENTIFIER || 'UNKNOWN_SYSTEMD_APP'
+      , f: data.CONTAINER_NAME || data._SYSTEMD_UNIT || data.SYSLOG_IDENTIFIER || 'UNKNOWN_SYSTEMD_APP'
```
darinspivey, line 226: For the record, I have figured out to get this tested using journald within Docker, but it needs polished.  It was a pretty painful process.

smusali: Gr8!

darinspivey (approved):

mdeltito (approved):

manigandham:
This change has made all log entries by the `logdna-agent` use `fluentbit.log` on our GKE kubernetes cluster. Our exclusion rules for `logdna-agent` as the app no longer work. 

Is this intended? Is there a way to set the agent name explicitly?

image

darinspivey:
Hi @manigandham .  What was the value of that field prior to the latest change?  I suppose this is expected behavior since the value of that field can be a [variety of values](https://github.com/logdna/logdna-agent/blob/19cbbf18d6998e7df6ae3b2e9d42f352cd7a7d2e/lib/file-utilities.js#L226).  The field is *meant* to be the filename of the file that contains the log message, so it seems correct.  What's the reason this is problematic for you?

In a future change, we will have to allow this field to be specifically set by the user, but it will have to be added as a feature.  We realize that until the above line is broken up into separate values configurable by the user, we will not be able to make all users happy (it's fixed for some users, but for others like yourself, it "broke" the expected value).

manigandham:
It was previously set to `logdna-agent` and now reads `fluentbit.log`

This is the `App` field so should it be the filename instead of what's producing the logs? Especially since there's already a `File` label. I don't have enough retention to show an example of a log line from before but this is a current log entry:

image

Or is this a parsing issue for the logs? Let me know if I'm off base here.

darinspivey:
Can you confirm for me what version you're using?  I see that the log levels are lower-cased (`debug`), but the most recent version uses upper-cased levels for consistency.

manigandham:
image is: `logdna/logdna-agent-v2:stable`

last pulled on 11/25

darinspivey:
Ah, ok, that tells us everything.  We're really sorry for the confusion, but the version you're running is actually our Rust  version, not the code on this PR.  We're working on renaming the two versions appropriately, but it's a long process to change legacy names, so thanks for bearing with us.

We need to move this topic off of this old PR and get you onto a new issue.  Would you mind opening an official issue under the "v2" repo describing what you're expecting to see?  The screen shots really helped too.  The correct repo is this: https://github.com/logdna/logdna-agent-v2/issues

Thanks!

manigandham:
Ok done, thanks. https://github.com/logdna/logdna-agent-v2/issues/51

Issue #212, opened 2020-11-10
Deploying on Elastic beanstalk bug


CarlChaappletree:
Hi,

Deploying app on EB, building fails with the following message:

Activity execution failed, because: Stopping logdna-agent.............
  Not stopped; may still be shutting down or shutdown may have failed
  Unable to stop, will not attempt to start
   (ElasticBeanstalk::ExternalInvocationError)

The weird thing is that first build fails but right after building fail, if I build the second with the same app, it works. so basically building two times in a role works.

Does anyone have an idea?

More info:
- It starts happening about 2 weeks ago.

- I use the following extension script as your docs say.
https://docs.logdna.com/docs/elastic-beanstalk

CarlChaappletree:
I use the following extension script: 

files:
  "/home/ec2-user/logdna.sh" :
    mode: "000777"
    owner: root
    group: root
    content: |
      #!/bin/sh
      rpm --import https://repo.logdna.com/logdna.gpg
      echo "[logdna]
      name=LogDNA packages
      baseurl=https://repo.logdna.com/el6/
      enabled=1
      gpgcheck=1
      gpgkey=https://repo.logdna.com/logdna.gpg" | sudo tee /etc/yum.repos.d/logdna.repo
      yum -y install logdna-agent
      logdna-agent -k INSERT_KEY_HERE # this is your unique Ingestion Key
      # /var/log is monitored/added by default (recursively), optionally add more dirs here (remove # to uncomment the line)
      #logdna-agent -d /path/to/your/logs
      # --hostname allows you to pass your env hostname/appname to LogDNA, optionally you can change this to whatever you want (remove # to uncomment the line below)
      #logdna-agent --hostname `{"Ref": "AWSEBEnvironmentName" }`
      # -t option allows you to tag the host/app with the existing host/app name, optionally you can change this to whatever you want (remove # to uncomment the line below)
      #logdna-agent -t `{"Ref": "AWSEBEnvironmentName" }`
      chkconfig logdna-agent on
      service logdna-agent start
commands:
  01_install_logdna:
    command: "/home/ec2-user/logdna.sh"
  02_restart_logdna:
    command: "service logdna-agent restart"

jalenplayvs:
[...]

PR #213, opened 2020-11-10
fix(init-script): Send SIGKILL if the process refuses to stop

jalenplayvs:
Works around #212, Elastic Beanstalk deployment failures.

Fixes: #212
Semver: patch

### DIFF
--- a/tools/files/linux/init-script
+++ b/tools/files/linux/init-script
@@ -28,8 +28,26 @@ get_pid() {
     cat "$pid_file"
 }
 
+rm_pid_file() {
+    if [ -f "$pid_file" ]; then
+        rm "$pid_file"
+    fi
+}
+
 is_running() {
     [ -f "$pid_file" ] && ps `get_pid` > /dev/null 2>&1
+}
+
+wait_for_exit() {
+    for i in {1..10}
+    do
+        if ! is_running; then
+            break
+        fi
+
+        echo -n "."
+        sleep 1
+    done
 }
 
 case "$1" in
@@ -58,26 +76,29 @@ case "$1" in
     if is_running; then
         echo -n "Stopping $name..."
         kill `get_pid`
-        for i in {1..10}
-        do
-            if ! is_running; then
-                break
-            fi
-
-            echo -n "."
-            sleep 1
-        done
+        wait_for_exit
         echo
 
         if is_running; then
-            echo "Not stopped; may still be shutting down or shutdown may have failed"
-            echo "Not stopped; may still be shutting down or shutdown may have failed" >> $stdout_log
-            exit 1
+            msg="Not stopped; may still be shutting down or shutdown may have failed, sending SIGKILL"
+            echo "$msg"
+            echo "$msg" >> $stdout_log
+            kill -9 `get_pid`
+            wait_for_exit
+            if is_running; then
+                msg="Failed to stop service"
+                echo "$msg"
+                echo "$msg" >> $stdout_log
+                exit 1
+            else
+                msg="Stopped using SIGKILL"
+                echo "$msg"
+                echo "$msg" >> $stdout_log
+                rm_pid_file
+            fi
         else
             echo "Stopped"
-            if [ -f "$pid_file" ]; then
-                rm "$pid_file"
-            fi
+            rm_pid_file
         fi
     else
         echo "Not running"


tools/files/linux/init-script
```diff
@@ -58,26 +76,29 @@ case "$1" in
...
         echo
 
         if is_running; then
-            echo "Not stopped; may still be shutting down or shutdown may have failed"
-            echo "Not stopped; may still be shutting down or shutdown may have failed" >> $stdout_log
-            exit 1
+            msg="Not stopped; may still be shutting down or shutdown may have failed, sending SIGKILL"
+            echo "$msg"
+            echo "$msg" >> $stdout_log
+            kill -9 `get_pid`
```
darinspivey, line 86: So with these changes, it fails to stop with the default `SIGTERM` signal?  When running this process with `node index.js`, the TERM signal properly shuts down, so we should understand why that doesn't work rather than KILL (if what I've suggested is the case)

darinspivey:
Thanks for the PR, @jalenplayvs , we will review it.  Can you fix your commit message, though?  We'd prefer the following format (sorry, we've got commit templates in flight but they haven't landed in this repo yet):
```
fix(init-script): [your title]
(blank line)
[your body was fine]

Fixes: [the issue link]
Semver: patch

```

tools/files/linux/init-script
jalenplayvs: > So with these changes, it fails to stop with the default SIGTERM signal

This is only invoked if SIGTERM fails to stop the process in 10 seconds

> When running this process with node index.js, the TERM signal properly shuts down, so we should understand why that doesn't work rather than KILL (if what I've suggested is the case)

Except that it does not always stop, as in the referenced issue #212 .  The 2 are not mutually exclusive, the root cause of the process refusing to stop should also be fixed.

I would argue that guaranteeing that `service logdna-agent stop` will always stop the process is a preferable strategy to always expecting `SIGTERM` to succeed (and causing production issues for some customers if it does not).

jalenplayvs: Also, `systemd` is the current standard for init systems, the default configuration for `systemd` services is:
```
KillSignal=SIGTERM
FinalKillSignal=SIGKILL
SendSIGKILL=yes
```
Which this PR roughly replicates that behavior.

Ref: https://www.freedesktop.org/software/systemd/man/systemd.kill.html

darinspivey: Yep, all that's good, and we agree that SIGKILL should be a fallback to use.  I'm happy to report that we've found the actual shutdown issue to be related to some timers that needed to be cleaned up on shutdown, and that's what's causing `SIGTERM` not to work.

We will need to open a separate PR to fix the underlying issue, but we think your PR is useful to allow a definite KILL of the process if need be.  We should have our fixes landed in the next few days, so feel free to pull those changes and continue with this PR to update the bash script.

jalenplayvs:
@darinspivey @c-nixon @jakedipity :  The other PR did not fully fix the issue for us, although it did seem to reduce the frequency.  We have over 10 environments being CI/CD'd, so the high throughput still results in a fairly large number of failures.  It appears that logdna-agent is eventually stopping now (some/most/all of the time?), but taking longer than 10 seconds for a small fraction of the deploys.

With the current init script, if the init script reports failure for a stop that eventually succeeds, the instance will continue running with no logs being sent LogDNA.  This PR would fix that, can we please get it reviewed, merged and released soon?  If we want to consider giving it more than 10 seconds to stop, let me know how long and I am happy to make that part of this PR.

darinspivey:
> This PR would fix that, can we please get it reviewed, merged and released

Well, that's a shame it didn't fully fix the issue.  We can certainly get the `kill` PR reviewed and merged at this point since kill should certainly be the fallback if a 10-second SIGTERM isn't working.

Let me ask one final question - how many log files are in any given instance?  10 seconds should be more than enough time to shut down dozens of log files, so increasing the time doesn't feel like it would help.

darinspivey (approved):
Thanks!

jalenplayvs:
> Let me ask one final question - how many log files are in any given instance? 10 seconds should be more than enough time to shut down dozens of log files, so increasing the time doesn't feel like it would help.

We have multiple applications sending to LogDNA, they seem to be around 20 logs each, give or take a few.  Perhaps a handful of those (at most) being high volume.

smusali (approved):
looks good to be merged!

smusali:
@jalenplayvs, this fix for #212 has already been released in the newest version: `v2.1.2` - thanks for the fix!

jalenplayvs:
Awesome, thanks everybody!

timbrownsf:
Thank you! We saw this popup once yesterday for some reason, but hasnt happened otherwise (and it was happening on every deploy prior). Will let you know if it rears its ugly head again. Thanks so much for the quick turn around!

PR #215, opened 2020-11-11
fix(TailReadStream): Cancel timers to prevent shutdown hanging

darinspivey:
There was an issue when running the agent as a service where the timers
in `TailReadStream` were hanging the node process. This might be a
difference between the `nexe` binary version and the plain version
(since this issue is not seen when running via node.js directly).
This change adds tracking for all timers in that code, and issues a
`cancelTimeout` for each of them during shutdown. The problem was
verified/reproduced and fixed using a Docker image of Debian.

NOTE: There is currently no test in the test suite that can verify this
change since it already shuts down properly there. Future versions of
the agent will, however, include proper testing and will not be using
`TailReadStream` for file consumption.

Semver: patch
Ref: LOG-7930
Fixes: https://github.com/logdna/logdna-agent/issues/212

### DIFF
--- a/lib/tailreadstream/tailreadstream.js
+++ b/lib/tailreadstream/tailreadstream.js
@@ -35,6 +35,10 @@ class TailReadStream extends Readable {
     this._paused = false
     this._ended = false
     this._watching = false
+    this._exiting = false
+    this._retryTimer = null
+    this._watchTimer = null
+    this._readTimer = null
   }
 
   static tail(filepath, fromstart, options) {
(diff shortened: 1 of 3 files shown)


lib/tailreadstream/tailreadstream.js
```diff
@@ -217,7 +226,8 @@ class TailReadStream extends Readable {
 
     fs.stat(this._filepath, (err, stats) => {
       if (err) {
-        return setTimeout(that._watchFile.bind(that), that._watchinterval)
+        this._watchTimer = setTimeout(that._watchFile.bind(that), that._watchinterval)
```
mdeltito, line 229: do we need a check for `this._exiting`? seems like there could be timing issues with how/when `_watchFile` and `destroy` get called, but not totally sure.

darinspivey: There's a small window of opportunity for a race condition, sure.  Couldn't hurt to add it.

mdeltito (approved):

smusali (approved):
I think, you pretty much cover all by adding `this._exiting` check; so, lgtm!

PR #219, opened 2021-01-19
deps: Bump @logdna/logger and respect sendUserAgent option

darinspivey:
**feat: Honor passthrough option for sendUserAgent**

The logger will respect the option for sendUserAgent, so
make sure that is passed through.

Semver: minor

---

**deps: @logdna/logger@2.2.0**

This bump contains additions to the Error metadata and a few
minor commits.

Semver: patch

### DIFF
--- a/index.js
+++ b/index.js
@@ -222,6 +222,7 @@ function loadConfig(program) {
       , ...parsedConfig
       , tags: process.env.LOGDNA_TAGS || parsedConfig.tags
       , rescanTimer: null
+      , sendUserAgent: program.sendUserAgent || process.env.LOGGER_SEND_USER_AGENT
       }
 
       if (config.exclude_regex) {
--- a/test/unit/logger-client.js
+++ b/test/unit/logger-client.js
@@ -1,0 +1,60 @@
+'use strict'
+
+// Testing env vars is painful here. Override the default and check config-driven
+// option later
+process.env.LOGGER_SEND_USER_AGENT = ''
+
+const {test} = require('tap')
+const loggerClient = require('../../lib/logger-client.js')
+const config = require('../../lib/config.js')
+
+const CONFIG_DEFAULT = {
+  ...config
+, key: 'abc123'
+, UserAgent: 'fake-transport@1.2.3'
+, hostname: 'myhostname'
+}
+
+test('Check logger defaults', async (t) => {
+  const logger = loggerClient.createLoggerClient(CONFIG_DEFAULT)
+  const requestDefaults = logger[Symbol.for('requestDefaults')]
+
+  t.match(requestDefaults.headers, {
+    'Content-Type': 'application/json; charset=UTF-8'
+  , 'Content-Encoding': 'gzip'
+  }, 'Default headers are correct')
+
+  const agentRegex = /@logdna\/logger\/\d+\.\d+\.\d+ \(fake-transport@1\.2\.3\)/
+  const header = logger[Symbol.for('userAgentHeader')]
+
+  t.equal(agentRegex.test(header), true, 'UserAgent header is correct')
+
+  t.equal(logger.sendUserAgent, true, 'sendUserAgent is taken from the env')
+})
+
+test('Check logger passthrough options for flushing', async (t) => {
+  const logger = loggerClient.createLoggerClient({
+    ...CONFIG_DEFAULT
+  , FLUSH_LIMIT: 12345
+  , FLUSH_INTERVAL: 34567
+  })
+  const requestDefaults = logger[Symbol.for('requestDefaults')]
+
+  t.match(requestDefaults.headers, {
+    'Content-Type': 'application/json; charset=UTF-8'
+  , 'Content-Encoding': 'gzip'
+  }, 'Default headers are correct')
+
+  t.equal(logger.flushLimit, 12345, 'flushLimit is passed through')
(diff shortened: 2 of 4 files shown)


index.js
```diff
@@ -205,6 +205,10 @@ function loadConfig(program) {
         saveMessages.push(`Tags: ${processed.diff} been saved to config.`)
       }
 
+      if (program.sendUserAgent || process.env.LOGGER_SEND_USER_AGENT) {
+        parsedConfig.sendUserAgent = program.sendUserAgent || process.env.LOGGER_SEND_USER_AGENT
```
smusali, line 209: @darinspivey I see [the default value](https://github.com/logdna/logger-node/blob/master/lib/logger.js#L79) for `sendUserAgent` within `logger-node` is `true` - I'd suggest to use it here too to make the default / fallback value visible - overall, I'd say instead of `if` block, my suggestion would be just
```
parsedConfig.sendUserAgent = program.sendUserAgent || process.env.LOGGER_SEND_USER_AGENT || true
```

smusali (commented):
just a small suggestion but the rest is good

index.js
darinspivey: Hmm, not sure I agree.  Doing this would set a default for the default.  I feel like that should be handled by the logger, therefore don't pass anything if it's not defined.

test/unit/logger-client.js
```diff
@@ -0,0 +1,67 @@
...
+  const logger = loggerClient.createLoggerClient(CONFIG_DEFAULT)
+  const requestDefaults = logger[Symbol.for('requestDefaults')]
+
+  t.match(requestDefaults.headers, {
+    'Content-Type': 'application/json; charset=UTF-8'
+  , 'Content-Encoding': 'gzip'
+  }, 'Default headers are correct')
+
+  const expectedAgentHeader = `@logdna/logger/${
+    pkg.dependencies['@logdna/logger'].replace('^', '')
```
mdeltito, line 29: technically the semver in package.json might differ from the version in the package-lock.json. probably a non-issue in practice, but in theory you _could_ instead swap the require to be `package-lock.json` and always get the true resolved version.

mdeltito (dismissed):
LGTM

index.js
darinspivey: I can remove the `if` block though, if that works for you?

test/unit/logger-client.js
darinspivey: Ah, good point.  I think just making it a regex would suffice.  It feels like making it match exact is introducing unnecessary complexity.

darinspivey: done

index.js
```diff
@@ -216,6 +216,9 @@ function loadConfig(program) {
         return
       }
 
+      // Do NOT write this to the config file. It's meant as a one-time override.
+      parsedConfig.sendUserAgent = program.sendUserAgent || process.env.LOGGER_SEND_USER_AGENT
```
smusali, line 220: @darinspivey, this is fine but why setting it to `parsedConfig.sendUserAgent` but not directly to `config.sendUserAgent`? Is there any specific reason?

Another question is how `program.sendUserAgent` is gonna get used? What is the key? I mean, for tags, we have `-t` or `--tag` but how can that information come from `program` arguments? Because agent isn't like the code library in which you can add parameters while initiating

If there is a way to pass this information thru cli arguments, can you please update the `README.md` too?

smusali (changes requested):
Q: how is this gonna get used besides thru env variable?

test/unit/logger-client.js
```diff
@@ -0,0 +1,64 @@
...
+  }, 'Default headers are correct')
+
+  const expectedAgentHeader = /@logdna\/logger\/\d+\.\d+\.\d+ (fake-transport@1\.2\.3)/
+  t.match(
+    logger
+  , {
+      [Symbol.for('userAgentHeader')]: expectedAgentHeader
+    }
+  , 'UserAgent header is correct'
+  )
```
mdeltito, line 34: I don't think `match` will work with symbols since they are not enumerable
```js
test('foo()', async (t) => {
  const k = Symbol('foo')
  const o = {[k]: true}
  t.match(o, {[k]: false}) // passes
})
```

index.js
darinspivey: Yep, I can move the setting to the `config` object creation below.  Good call.

`program.sendUserAgent` is the "config-driven" approach that would be used in testing.  I'm using it in the test that was just made for this where it passed it directly.  The goal is to not have to use the env var because testing with those is a one-shot deal.

I need to update the README, thanks.

darinspivey: The README has no information about the env vars.  This is one of the things that would completely need re-worked.  If I add this, then we'd have to add all of the env vars, and I think that's out of scope.  I think we should decide as a group whether or not we should spend time there.

test/unit/logger-client.js
darinspivey: Hrm, well doing a `match` on the string value gave me plenty of problems.  Not sure what to do here.

darinspivey: fixed with a normal regex `.test`

mdeltito (approved):

index.js
smusali: Thanks, @darinspivey! It all makes sense

smusali (approved):
lgtm!

PR #229, opened 2021-12-17
Update README.md

braxtonj:
added manual install instructions for customers


braxtonj:
one more thing to add actually

PR #233, opened 2022-05-04
LOG-12515: Upgrade of code for Node 18

arani-sen:
Node 12 is deprecated so these changes are for continue use in Node 18. 

Ticket Ref: [LOG-12515](https://logdna.atlassian.net/browse/LOG-12515)


dkhokhlov (commented):
we might want to consider to use node v16 instead v18 as v16 is "stable" now.

index.js
```diff
@@ -29,34 +29,57 @@ process.title = 'logdna-agent'
 commander._name = 'logdna-agent'
 commander
   .version(pkg.version, '-v, --version')
-  .description('This agent collect and ship logs for processing. Defaults to /var/log if run without parameters.')
-  .option('-c, --config <file>', `uses alternate config file (default: ${config.DEFAULT_CONF_FILE})`)
+  .description(
+    'This agent collect and ship logs for processing.'
+      + 'Defaults to /var/log if run without parameters.'
```
darinspivey, line 34: Needs a space before `Defaults`

darinspivey: should be `collects and ships logs`, no?

index.js
```diff
@@ -29,34 +29,57 @@ process.title = 'logdna-agent'
...
+  )
+  .option(
+    '-c, --config <file>'
+  , `uses alternate config file (default: ${config.DEFAULT_CONF_FILE})`
```
darinspivey, line 38: Let's be consistent with capitalization.  Some of these sound like sentences, so I'm a fan of first letter capital.

darinspivey: btw, I know you didn't write any of these :)

lib/config.js
```diff
@@ -5,24 +5,35 @@ const os = require('os')
 const path = require('path')
 
 module.exports = {
-  AWS_INSTANCE_CHECK_URL: 'http://169.254.169.254/latest/dynamic/instance-identity/document/'
+  AWS_INSTANCE_CHECK_URL:
+  'http://169.254.169.254/latest/dynamic/instance-identity/document/'
```
darinspivey, line 9: Looks like this needs indenting?  Does the linter let you do that?

lib/file-utilities.js
```diff
@@ -102,64 +104,50 @@ function getFiles(config, dir, callback) {
...
-      tail.on('rename', () => {
-        log(`log rotated: ${file} by rename`)
-      })
-
-      tail.on('truncate', () => {
-        log(`log rotated: ${file} by truncation`)
-      })
-
       tailStreams.push(tail)
+    } catch (err) {
```
darinspivey, line 145: This doesn't actually work the way you intend because `.start()` returns a promise which needs to be `await`ed before an error would throw.  You can't do that here because we're not in an `async` function.  . Data also won't flow until you pipe, so you can set up your data listener first.  The correct way is to use thenables all throughout:

```js
tail.on('data', ...)
tail
  .start()
  .then(() => {
    if (callback) callback()
  })
  .catch((err) => {
     log.error(err, 'Tail could not start. Does the file exist?')
  })
```

lib/file-utilities.js
```diff
@@ -169,20 +157,25 @@ function streamAllLogs(config, callback) {
   async.each(config.logdir, (dir, done) => {
     getFiles(config, dir, (err, logfiles) => {
       if (!err && logfiles.length > 0) {
+        conf = config
```
darinspivey, line 160: Ugh, I hate that I never got to finish this.  `conf = config` is terrible, but oh well.

lib/file-utilities.js
```diff
@@ -270,18 +248,40 @@ function streamAllLogs(config, callback) {
     streamAllLogs(config)
   }, config.RESCAN_INTERVAL) // rescan for files every once in awhile
 }
-
-function gracefulShutdown(signal) {
+function processChunk(data) {
```
darinspivey, line 251: Add a blank line above this

lib/file-utilities.js
```diff
@@ -270,18 +248,40 @@ function streamAllLogs(config, callback) {
...
+  try {
+    data = JSON.parse(data)
```
darinspivey, line 253: We should avoid mutating inputs whenever possible.  It makes it confusing for no reason (re-using the var provides little efficiency).  Here, just change the parameter name to `input` and declare a new var.  That way the error can return `input` and it's clear to the reader that the parse failed at that point.

```js
function processChunk(input) {

  let data
  try {
     data = JSON.parse(input)
  } catch (err) {
    return {t: Date.now(), l: input, f: 'systemd'}
  }
```

tools/scripts/darwin.sh
```diff
@@ -5,7 +5,7 @@
 ARCH=x64
 INPUT_TYPE=dir
 LICENSE=MIT
-NODE_VERSION=12.16.2
+NODE_VERSION=14.15.3
```
darinspivey, line 8: LTS is now version 16.  Is there any reason why we're stopping at 14?

darinspivey (commented):
Looking good, just a few things.  Thanks!

darinspivey:
You should also squash your commits.  Sometimes we do that when merging, but I don't see a good commit message (in the right format) that we could squash into.  Perhaps just do it in dev and push up a single commit with these changes.  They're all related to the same "fix/upgrade", so it should be 1 commit anyway.

Start it with something like `fix(pkg): Upgrade to node xxxx` etc.

tools/scripts/darwin.sh
arani-sen: We are doing 14 because of the `nexe`. This is the highest/most common version of prebuilt builds of node. It just makes the creation easier/faster.

index.js
arani-sen: Yeah, that makes more sense.

index.js
arani-sen: Thanks for the look out. I know there were some legacy things so I appreciate you going through and seeing where things are wrong.

lib/file-utilities.js
arani-sen: Yeah I had my issues with that too. I just left it as is for the sake of not messing with this too much. The thought was this will eventually be deprecated.

index.js
```diff
@@ -29,34 +29,57 @@ process.title = 'logdna-agent'
...
-    console.log('    $ logdna-agent -u tags,logdir                                      # unset specific entries from config')
-    console.log('    $ logdna-agent -u all                                              # unset all except LogDNA API Key')
+    console.log('    $ logdna-agent -l tags,key,logdir                                  '
+      + '# show specific config parameters')
+    console.log('    $ logdna-agent -l                                                  '
+      + '# show all')
+    console.log('    $ logdna-agent -u tags,logdir                                      '
+      + '# unset specific entries from config')
+    console.log('    $ logdna-agent -u all                                              '
+      + '# unset all except LogDNA API Key')
```
esatterwhite, line 82: This is a bit hard to read and will be hard to maintain. You could use a single multi-line string template
```javascript
`
  this can contain 
  text on multiple lines
`.trim()
```

that can be a little hard to keep spacing, so the other option would be to put the help text formatted cleanly in a separate file and just console.log the contents of the file.

index.js
```diff
@@ -260,15 +290,18 @@ function loadConfig(program) {
...
+      .reduce((networkData, interfaces) => {
+        interfaces.forEach((interfce) => { return networkData.push(interfce) })
+        return networkData
+      }, []).filter((interfce) => {
+        return interfce.family
+          && interfce.family === 'IPv4'
+          && interfce.mac && interfce.address && (
+          interfce.address.startsWith('10.')
+            || interfce.address.startsWith('172.')
+            || interfce.address.startsWith('192.168.'))
```
esatterwhite, line 303: It a bit easier to read if all of the operators are on the same size

```javascript
        return interfce.family
          && interfce.family === 'IPv4'
          && interfce.mac 
          && interfce.address 
          && (
             interfce.address.startsWith('10.')
               || interfce.address.startsWith('172.')
               || interfce.address.startsWith('192.168.')
          )
```

index.js
```diff
@@ -287,14 +320,19 @@ function loadConfig(program) {
...
   commander.parse(process.argv)
-  const isAdmin = os.platform() === 'win32' ? require('is-administrator')() : process.getuid() === 0
+  const isAdmin = os.platform() === 'win32'
+    ? require('is-administrator')()
+    : process.getuid() === 0
 
   if (!isAdmin) {
-    console.log('You must be an Administrator (root, sudo) to run this agent! See -h or --help for more info.')
+    console.log('You must be an Administrator (root, sudo) to run this agent! '
+      + 'See -h or --help for more info.')
```
esatterwhite, line 335: before the return you ought to set a non-zero exit code
```javascript
process.exitCode = 1
```

lib/utils.js
```diff
@@ -102,12 +107,20 @@ function pick2list(options, config) {
...
+    return split(option, ',', hasWhitespace)
+  }))
+  const oldValues = (config
+    ? (typeof config === 'string'
+      ? split(config, ',', false)
+      : config)
+    : [])
+  const diff = newValues.filter((value) => {
+    return oldValues.indexOf(value) < 0
+  }).filter((value) => { return value })
```
esatterwhite, line 120: I think this is the same thing as `.filter(Boolean)` 

what are we trying to filter out here? empty strings?

lib/utils.js
```diff
@@ -162,7 +177,9 @@ function unsetConfig(options, config) {
     return obj
   }, {})
   const newValues = (config ? Object.keys(config) : [])
-  const diff = oldValues.filter(value => newValues.indexOf(value) < 0).filter(value => value)
+  const diff = oldValues.filter((value) => {
+    return newValues.indexOf(value) < 0
+  }).filter((value) => { return value })
```
esatterwhite, line 182: Could also potentially be `.filter(Boolean)` if we are just trying to remove falsy values

package.json
```diff
@@ -19,35 +19,34 @@
...
     "properties": "^1.2.1",
-    "request": "^2.88.2"
+    "request": "^2.88.2",
+    "split2": "^4.1.0"
   },
   "devDependencies": {
-    "eslint": "^7.6.0",
-    "eslint-config-logdna": "^1.0.0",
+    "eslint": "^7.32.0",
+    "eslint-config-logdna": "^5.1.0",
```
esatterwhite, line 49: Why 5.1 specifically? the latest is 6.1

lib/file-utilities.js
```diff
@@ -271,17 +257,40 @@ function streamAllLogs(config, callback) {
   }, config.RESCAN_INTERVAL) // rescan for files every once in awhile
 }
 
-function gracefulShutdown(signal) {
+function processChunk(input) {
+  try {
+    data = JSON.parse(input)
```
dkhokhlov, line 262: undefined `data` ?

index.js
```diff
@@ -99,11 +123,15 @@ function loadConfig(program) {
 
       if (!program.hostname && !parsedConfig.hostname) {
         if (fs.existsSync(HOSTNAME_PATH) && fs.statSync(HOSTNAME_PATH).isFile()) {
-          parsedConfig.hostname = fs.readFileSync(HOSTNAME_PATH).toString().trim().replace(HOSTNAME_IP_REGEX, '')
+          parsedConfig.hostname = fs.readFileSync(HOSTNAME_PATH)
+            .toString()
```
esatterwhite, line 127: You can pass encode parameter to `readFIleSync` so you don't have to run to string in process

```javascript
hostname = fs.readFIleSync(HOSTNAME_PATH, 'utf8').trim()
```

README.md
```diff
@@ -1,7 +1,8 @@
 # LogDNA Agent
 
 LogDNA Agent streams from log files to your LogDNA account. Works with Linux, Windows, and macOS Servers.
-
+## Deprecation Warning!!!
```
esatterwhite, line 4: More than one exclamaition point is bad punctuation. it should really only ever be one.
maybe using `:warning:`


## :warning: DEPRECATION WARNING

lib/file-utilities.js
```diff
@@ -102,64 +104,58 @@ function getFiles(config, dir, callback) {
...
-      tail.pipe(new Splitter())
+    tail.on('truncate', () => {
+      log(`log rotated: ${file} by truncation`)
+    })
+
+    try {
+      tail
+        .start()
+        .then(()=>{
+          if(callback) callback()
```
esatterwhite, line 138: I don't think the if check here is actually necessary

```javascript
tail
  .start()
  .then(callback)
  .catch((err) => {
    ...
  })
```

lib/file-utilities.js
```diff
@@ -205,35 +206,20 @@ function streamAllLogs(config, callback) {
...
-      }
-    }
-
     journalctl.stdout.on('data', (data) => {
       data = data.toString().trim().split('\n')
       for (i = 0; i < data.length; i++) {
-        if (data[i].substring(0, 1) === '{' && data[i].substring(data[i].length - 1) === '}') {
+        if (data[i].substring(0, 1)
+          === '{' && data[i].substring(data[i].length - 1)
+          === '}') {
```
esatterwhite, line 222: nit picky, but put each conditional on its own line. when its split like this it can be difficult to understand where one check starts and one check stops.

```javascript
if (
  data[i].substring(0, 1)=== '{' 
  && data[i].substring(data[i].length - 1)=== '}'
) {
```

 it would also make it a bit more obvious to pull the data[i] bits out in to a variable

```javascript
const start = data[i].substring(0, 1)
const end = data[i].substring(data[i].length - 1)

if (start === '{' && end === '}') {
  ...
}
```

lib/utils.js
```diff
@@ -53,20 +53,25 @@ function appender(xs) {
     return xs
   }
 }
-
+function padStart(number, expectedLength) {
+  const strNumber = String(Math.abs(number)).padStart(expectedLength, '0')
+  return (number < 0) ? `-${strNumber}` : strNumber
+}
```
jsumners, line 59: This can be reduced to (given that times are not negative):

```js
function padStart(number, len) {
  return `${number}`.padStart(len, '0')
}
```

lib/utils.js
```diff
@@ -53,20 +53,25 @@ function appender(xs) {
...
-    return (number < 0) ? `-${strNumber}` : strNumber
-  }
-  const date = `${dateObject.getFullYear()}-${padStart(dateObject.getMonth() + 1, 2)}-${padStart(dateObject.getDate(), 2)}`
-  const time = `${padStart(dateObject.getHours(), 2)}:${padStart(dateObject.getMinutes(), 2)}:${padStart(dateObject.getSeconds(), 2)}`
+  const date = `${dateObject.getFullYear()}-${
+    padStart(dateObject.getMonth() + 1, 2)
+  }-${padStart(dateObject.getDate(), 2)}`
+  const time = `${padStart(dateObject.getHours(), 2)}:${
+    padStart(dateObject.getMinutes(), 2)
+  }:${padStart(dateObject.getSeconds(), 2)}`
```
jsumners, line 68: I'd just add `luxon` as a dependency and use `.toFormat` https://moment.github.io/luxon/#/formatting?id=toformat. It'd be a lot easier to read and understand.

arani-sen: It's legacy code and this is just a format change. I could just add that. I am not trying to put too much into this as this repo will eventually be deprecated

lib/utils.js
arani-sen: I believe the intention was to remove empty strings.

index.js
arani-sen: The issue I had was that this is a single line text and eslint had an issue with the character count. If it with a string template, it will print out the new line as well. That's why I didn't go with it at first.

arani-sen: I know i can escape the new line but the formatting is off. I can adjust the other ones to match

lib/utils.js
jsumners: I am not blocking anything, but I think it is a good idea to consider the duration of "eventually deprecated", how long that deprecation will last, and if the tool will be maintained during that deprecation period. In my opinion, I think easier to maintain code would be better.

jsumners (dismissed):

lib/utils.js
darinspivey: I'm afraid to say that this is going to be around a while.  There are customers that will never upgrade, so we should find that balance between changing too much and doing things that'll help us maintain it.  AFAIK, we'd only be bumping node and patching critical dependency updates when they come up.

PR #242, opened 2023-05-17
fix(ci): gate builds from contributors outside of org

esatterwhite:
To prevent builds from auto running on every pull request which has to potential to execute malicious code and expose secrets, we will require that a maintainer from within the org explicitly trigger a build.


darinspivey (approved):
nice update

Issue #247, opened 2025-02-26
Regular expression injection


odaysec:
https://github.com/logdna/logdna-agent/blob/2313b15a8223905bbf69860ecd61eda79a866168/index.js#L279-L279

Constructing a regular expression with unsanitized user input is dangerous as a malicious user may be able to modify the meaning of the expression. In particular, such a user may be able to provide a regular expression fragment that takes exponential time in the worst case, and use that to perform a Denial of Service attack.

## Recommendation
Before embedding user input into a regular expression, use a sanitization function such as lodash's `_.escapeRegExp` to escape meta-characters that have special meaning.

## POC
The following shows a HTTP request parameter that is used to construct a regular expression without sanitizing it first:
```
var express = require('express');
var app = express();

app.get('/findKey', function(req, res) {
  var key = req.param("key"), input = req.param("input");

  // BAD: Unsanitized user input is used to construct a regular expression
  var re = new RegExp("\\b" + key + "=(.*)\n");
});
```
Instead, the request parameter should be sanitized first, for example using the function `_.escapeRegExp` from the lodash package. This ensures that the user cannot insert characters which have a special meaning in regular expressions.
```js
var express = require('express');
var _ = require('lodash');
var app = express();

app.get('/findKey', function(req, res) {
  var key = req.param("key"), input = req.param("input");

  // GOOD: User input is sanitized before constructing the regex
  var safeKey = _.escapeRegExp(key);
  var re = new RegExp("\\b" + safeKey + "=(.*)\n");
});
```
## References
OWASP: [Regular expression Denial of Service - ReDoS](https://www.owasp.org/index.php/Regular_expression_Denial_of_Service_-_ReDoS)
Wikipedia: [ReDoS](https://en.wikipedia.org/wiki/ReDoS)
npm: [lodash](https://www.npmjs.com/package/lodash)
Common Weakness Enumeration: [CWE-730](https://cwe.mitre.org/data/definitions/730.html)
Common Weakness Enumeration: [CWE-400](https://cwe.mitre.org/data/definitions/400.html)

c-nixon:
Thanks for the report, this repository is no longer maintained.

kreeksec:
hi @c-nixon can you merged this `pull-request` as for fix this issue https://github.com/logdna/logdna-agent/pull/248

c-nixon:
[...]

PR #248, opened 2025-02-26
fix: Regular expression injection

odaysec:
fix vulnerable in #247

To fix the problem, we need to sanitize the user input before using it to construct the regular expression. The best way to do this is by using the `_.escapeRegExp` function from the lodash library, which escapes special characters in the input string. This ensures that the user cannot insert characters that have special meaning in regular expressions.

1. Install the lodash package if it is not already installed.
1. Import the lodash library in the file.
1. Use the `_.escapeRegExp` function to sanitize the `config.exclude_regex` before constructing the regular expression.

Suggested fixes powered by [Zeroday Operation Research LLC](https://github.com/zerodayresearch). 

## References
OWASP: [Regular expression Denial of Service - ReDoS](https://www.owasp.org/index.php/Regular_expression_Denial_of_Service_-_ReDoS)
Wikipedia: [ReDoS](https://en.wikipedia.org/wiki/ReDoS)
npm: [lodash](https://www.npmjs.com/package/lodash)
Common Weakness Enumeration: [CWE-730](https://cwe.mitre.org/data/definitions/730.html)
Common Weakness Enumeration: [CWE-400](https://cwe.mitre.org/data/definitions/400.html)


c-nixon:
Thanks for the PR, this repository is no longer maintained.

